PR #5, opened 2021-05-02
Make sure we don't have anyone called devbot

TommyPujol06:
Very simple commit to make nobody spoofs the `devbot` user.


quackduck:
Yep thanks. I was initially thinking of using a special color combo with the devbot user.

PR #6, opened 2021-05-06
updated readme

Mudrank:
more changes coming


quackduck:
I like the credit idea!

Mudrank:
Sure @quackduck! , agree

PR #15, opened 2021-05-07
Update README.md

IamTregsthedev:


IamTregsthedev:
@quackduck added readme change + elon

PR #20, opened 2021-05-19
Make tic-tac-toe interface cleaner

TommyPujol06:
Automate the printing of the board and add numbers to the cell number, so it's easier for the user.

Image of new board: https://i.ibb.co/xGF5pnS/pic-selected-210519-2033-34.png

I also added this message, so the user knows what to do.
> devbot: To make your move type /tic <move>

### DIFF
--- a/devchat.go
+++ b/devchat.go
@@ -80,7 +80,7 @@ var (
 
 	tttGame       = new(ttt.Board)
 	currentPlayer = ttt.X
-
+	hasStartedGame = false
 	hangGame = new(hangman)
 
 	admins = []string{"d84447e08901391eb36aa8e6d9372b548af55bee3799cd3abb6cdd503fdf2d82", "f5c7f9826b6e143f6e9c3920767680f503f259570f121138b2465bb2b052a85d", "6056734cc4d9fce31569167735e4808382004629a2d7fe6cb486e663714452fc"}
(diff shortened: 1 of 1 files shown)


quackduck:
I've tried the numbers thing, see the comments above your changes to the tttPrint func. They look good at the start of the game but then when people start playing, the "O" and "X" don't look so nice. So I even tried to use unicode subscript numbers, but that didn't look great either.

I like the changed message, but the user still doesn't know what a "<move>" actually is, so could you make the message "Play using /tic <cell num>", which is short and clear enough.

Thanks for your help with Devzat!

TommyPujol06:
Ok, I'll try to fix that today.

quackduck:
Dunno if it rendered correctly in the mail but I meant the improved message to have `<cell num`. (md interpreted it as html angle brackets)

quackduck:
Could you instead check if the game is new by checking if the array of cells is made of empty cells? Also, let the code to actually do the printing remain:
```go
	var buf bytes.Buffer
	fmt.Fprintf(&buf, " %v │ %v │ %v \n", strcells[0], strcells[1], strcells[2])
	fmt.Fprintln(&buf, "───┼───┼───")
	fmt.Fprintf(&buf, " %v │ %v │ %v \n", strcells[3], strcells[4], strcells[5])
	fmt.Fprintln(&buf, "───┼───┼───")
	fmt.Fprintf(&buf, " %v │ %v │ %v ", strcells[6], strcells[7], strcells[8])
```

TommyPujol06:
I messed up something with git and now there are many commits in here. I'm currently trying to fix it.

TommyPujol06:
OK, the issue has been fixed.

quackduck:
No problem

quackduck:
Love it! Thanks for helping Devzat!

PR #22, opened 2021-06-05
adding rainbow as secret color

ciearius:
restructured u.color and color management to encourage easier special color adding

### DIFF
--- a/devchat.go
+++ b/devchat.go
@@ -69,41 +69,19 @@ var (
 		{"easter", buildStyleApplicator(color.New(color.BgMagenta, color.FgHiYellow))},
 		{"baby", buildStyleApplicator(color.New(color.BgBlue, color.FgHiMagenta))},
 		{"hacker", buildStyleApplicator(color.New(color.FgHiGreen, color.BgBlack))},
-		//TODO: find a way to support special colors which rely on the user object
-		// {
-		// 	"l33t",
-		// 	func(a string, u user) string {
-		//		// TODO: this will lead to catastrophic failure with other special colors!
-		// 		c := *u.getColor().Add(color.BgHiBlack)
-		// 		return c.Sprint(stripansi.Strip(a))
-		// 	},
-		// },
-		// Same with "whiten", *u.getColor().Add(color.BgWhite)
-		{
-			"rainbow",
-			func(a string) string {
-				var rainbow = []*color.Color{
-					red,
-					magenta,
-					blue,
-					cyan,
-					yellow,
-					green,
-					white,
-				}
-
-				var stripped = stripansi.Strip(a)
-				var buf = ""
-				colorOffset := rand.Intn(len(rainbow))
-
-				for i, s := range stripped {
-					colorIndex := (colorOffset + i) % len(rainbow)
-					buf += rainbow[colorIndex].Sprint(string(rune(s)))
-				}
-
-				return buf
-			},
-		},
+		{"l33t", func(s string) string { return color.New(color.BgHiBlack).Sprint(s) }},
+		{"whiten", func(s string) string { return color.New(color.BgWhite).Sprint(s) }}, // TODO: remove extra reset ANSI code at the end
+		{"rainbow", func(a string) string {
+			var rainbow = []*color.Color{red, magenta, blue, cyan, yellow, green, white}
+			a = stripansi.Strip(a)
+			var buf = ""
+			colorOffset := rand.Intn(len(rainbow))
+			for i := range []rune(a) {
+				colorIndex := (colorOffset + i) % len(rainbow)
+				buf += rainbow[colorIndex].Sprint(string(a[i]))
+			}
+			return buf
+		}},
 	}
 
 	mainRoom = &room{"#main", make([]*user, 0, 10), sync.Mutex{}}
@@ -178,7 +156,7 @@ func main() {
 		}
 	}
 
-	fmt.Println(fmt.Sprintf("Starting chat server on port %d", port))
+	fmt.Printf("Starting chat server on port %d\n", port)
 	go getMsgsFromSlack()
 	go func() {
 		if port == 22 {
@@ -501,11 +479,9 @@ func (u *user) pickUsername(possibleName string) {
 // Applies color from name
 func (u *user) changeColor(colorName string) {
 	color := getColor(colorName)
-	u.changeColorstyle(color)
-}
-
-// Applies color from colorstyle
-func (u *user) changeColorstyle(color *colorstyle) {
+	if color == nil {
+		return
+	}
 	u.color = color.name
 	u.name = color.apply(u.name)
 	u.term.SetPrompt(u.name + ": ")
@@ -522,7 +498,7 @@ func getColor(name string) *colorstyle {
 			return colorstyles[i]
 		}
 	}
-	return colorstyles[0]
+	return nil
 }
 
 func (u *user) changeRoom(r *room, toSlack bool) {
--- a/devchat.go
+++ b/devchat.go
@@ -138,10 +138,10 @@ var (
 		"12a9f108e7420460864de3d46610f722e69c80b2ac2fb1e2ada34aa952bbd73e"} // jmw: github.com/ciearius
 )
 
-func buildStyleApplicator(c *color.Color) *(func(string) string) {
-	return &(func(s string) string {
+func buildStyleApplicator(c *color.Color) func(string) string {
+	return func(s string) string {
 		return c.Sprint(stripansi.Strip(s))
-	})
+	}
 }
 
 // TODO: have a web dashboard that shows logs
@@ -505,7 +505,7 @@ func (u *user) changeColor(colorName string) {
 }
 
 // Applies color from colorstyle
-func (u *user) changeColorstyle(color colorstyle) {
+func (u *user) changeColorstyle(color *colorstyle) {
 	u.color = color.name
 	u.name = color.apply(u.name)
 	u.term.SetPrompt(u.name + ": ")
@@ -1266,5 +1266,5 @@ func remove(s []*user, a *user) []*user {
 
 type colorstyle struct {
 	name  string
-	apply *(func(string) string)
+	apply func(string) string
 }
--- a/devchat.go
+++ b/devchat.go
@@ -54,10 +54,57 @@ var (
 	blue        = color.New(color.FgHiBlue)
 	black       = color.New(color.FgBlack)
 	white       = color.New(color.FgHiWhite)
-	colorArr    = []*color.Color{yellow, cyan, magenta, green, white, blue, red}
-	colorNames  = []string{"yellow", "cyan", "magenta", "green", "white", "blue", "red"}
 	devbot      = "" // initialized in main
 	startupTime = time.Now()
+
+	colorstyles = []colorstyle{
+		{"white", buildStyleApplicator(*white)},
+		{"red", buildStyleApplicator(*red)},
+		{"green", buildStyleApplicator(*green)},
+		{"cyan", buildStyleApplicator(*cyan)},
+		{"magenta", buildStyleApplicator(*magenta)},
+		{"yellow", buildStyleApplicator(*yellow)},
+		{"blue", buildStyleApplicator(*blue)},
+		{"black", buildStyleApplicator(*black)},
+		{"easter", buildStyleApplicator(*color.New(color.BgMagenta, color.FgHiYellow))},
+		{"baby", buildStyleApplicator(*color.New(color.BgBlue, color.FgHiMagenta))},
+		{"hacker", buildStyleApplicator(*color.New(color.FgHiGreen, color.BgBlack))},
+		//TODO: find a way to support special colors which rely on the user object
+		// {
+		// 	"l33t",
+		// 	func(a string, u user) string {
+		//		// TODO: this will lead to catastrophic failure with other special colors!
+		// 		c := *u.getColor().Add(color.BgHiBlack)
+		// 		return c.Sprint(stripansi.Strip(a))
+		// 	},
+		// },
+		// Same with "whiten", *u.getColor().Add(color.BgWhite)
+		{
+			"rainbow",
+			func(a string) string {
+				var rainbow = []*color.Color{
+					red,
+					magenta,
+					blue,
+					cyan,
+					yellow,
+					green,
+					white,
+				}
+
+				var stripped = stripansi.Strip(a)
+				var buf = ""
+				colorOffset := rand.Intn(len(rainbow) - 1)
+
+				for i, s := range stripped {
+					colorIndex := (colorOffset + i) % len(rainbow)
+					buf += rainbow[colorIndex].Sprint(string(rune(s)))
+				}
+
+				return buf
+			},
+		},
+	}
 
 	mainRoom = &room{"#main", make([]*user, 0, 10), sync.Mutex{}}
 	rooms    = map[string]*room{mainRoom.name: mainRoom}
@@ -89,6 +136,12 @@ var (
 		"e9d47bb4522345d019086d0ed48da8ce491a491923a44c59fd6bfffe6ea73317", // Arav Narula
 		"1eab2de20e41abed903ab2f22e7ff56dc059666dbe2ebbce07a8afeece8d0424"} // Shok
 )
+
+func buildStyleApplicator(c color.Color) styleApplication {
+	return func(s string) string {
+		return c.Sprint(stripansi.Strip(s))
+	}
+}
 
 // TODO: email people on ping word idea
 func main() {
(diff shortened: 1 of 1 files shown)


ciearius:
@quackduck

quackduck:
Sorry for not responding! I’m gonna look at this in a few hours, sorry! 

Thank you so much for contributing to Devzat!

devchat.go
```diff
@@ -437,12 +438,52 @@ func (u *user) pickUsername(possibleName string) {
...
+		if colorNames[i] == u.color {
+			return colorArr[i]
+		}
+	}
+	return colorArr[0]
+}
+
+func (u *user) rainbowColor() {
+	var rainbow = [...]color.Color{
+		*green,
```
quackduck, line 466: (about line 466) If I remember correctly, these (green, yellow, etc.) are already pointers to a color, so it might be a good idea to not take a pointer of them.

devchat.go
```diff
@@ -437,12 +438,52 @@ func (u *user) pickUsername(possibleName string) {
...
+	}
+
+	var stripped = stripansi.Strip(u.name)
+	var buf = ""
+	colorOffset := rand.Intn(len(rainbow) - 1)
+
+	for i, s := range stripped {
+		colorIndex := (colorOffset + i) % len(rainbow)
+		buf += rainbow[colorIndex].Sprint(string(rune(s)))
+		println(colorIndex, string(rune(s)))
```
quackduck, line 482: Does this code compile? No such function as println IIRC

quackduck: Were you using the `Println()` function to test out your code?

devchat.go
```diff
@@ -437,12 +438,52 @@ func (u *user) pickUsername(possibleName string) {
...
-	u.color = color
+	u.color = colorName
+	u.term.SetPrompt(u.name + ": ")
+	allUsersMutex.Lock()
+	allUsers[u.id] = u.name
+	allUsersMutex.Unlock()
+	saveBansAndUsers()
+}
+
+func (u *user) getColor() *color.Color {
```
quackduck, line 455: Might be a good idea to make a global variable that maps colors as strings to the actual color. (could replace this entire function)

devchat.go
```diff
@@ -183,7 +184,7 @@ type user struct {
 	session       ssh.Session
 	term          *terminal.Terminal
 	bell          bool
-	color         color.Color
+	color         string
```
quackduck, line 187: I don't really understand the motivation behind changing this to a string, what did you have in mind?

devchat.go
```diff
@@ -437,12 +438,52 @@ func (u *user) pickUsername(possibleName string) {
...
+		*yellow,
+		*white,
+	}
+
+	var stripped = stripansi.Strip(u.name)
+	var buf = ""
+	colorOffset := rand.Intn(len(rainbow) - 1)
+
+	for i, s := range stripped {
+		colorIndex := (colorOffset + i) % len(rainbow)
```
quackduck, line 480: this is really beautiful code man

quackduck: lol

quackduck (changes requested):
Do let me know the motivation behind the new array of strings!

devchat.go
ciearius:
```suggestion
```
Yes, I forgot to remove my debug statement.

quackduck: Ah no problem lol

quackduck:
hey @ciearius how's it goin

devchat.go
ciearius: The point is, to support colors beyond single colored strings. Other colors would still be possible with `u.color` being a `color.Color`, but we can not store the state of those special colors then!

quackduck: ah so you want it to be able to be serialized?

ciearius: Yes! Absolutely. We may need this later, right? And as rainbow is not storable in `color.Color`. Afaik a string or an enum would be the way to go... Do you have another way?

quackduck: Yep that looks pretty nice to me (separating the two changes into two prs would have been nice but this is fine! we're gonna serialize more data (such as bell settings, timezone) later on anyway)

devchat.go
```diff
@@ -45,16 +46,16 @@ var (
...
-
+	red         = color.New(color.FgHiRed)
+	green       = color.New(color.FgHiGreen)
+	cyan        = color.New(color.FgHiCyan)
+	magenta     = color.New(color.FgHiMagenta)
+	yellow      = color.New(color.FgHiYellow)
+	blue        = color.New(color.FgHiBlue)
+	black       = color.New(color.FgBlack)
+	white       = color.New(color.FgHiWhite)
+	colorArr    = []*color.Color{yellow, cyan, magenta, green, white, blue, red}
```
quackduck, line 57: okie so we can rename this variable to `colorMap` and make it a `map[string]*color.Color`

devchat.go
```diff
@@ -45,16 +46,16 @@ var (
...
+	colorNames  = []string{"yellow", "cyan", "magenta", "green", "white", "blue", "red"}
```
quackduck, line 58: Since we're making a map this variable is no longer needed

devchat.go
ciearius: Good point. Removed the references!

quackduck:
Actually wait we should not make a map

devchat.go
quackduck: Thanks!

quackduck:
We should make a struct called `color` that has two fields:
1. `Name string`: stores the name of the color in a string
2. `Apply func(string) string`: a function that prints the input string in the color (this is so rainbow can also be saved as a normal `color`.

ciearius:
Very good idea!

quackduck:
Let me know if you think that it's a better idea to instead make your array of names an array of `func(string) string`s instead of making a new struct!

quackduck (changes requested):

devchat.go
quackduck: Nope sorry, this will have to be thought through better

devchat.go
quackduck: nope, I'm wrong

devchat.go
quackduck: nope that's a bad idea :P

quackduck:
The reason we're not using a map is because picking random values from maps is hard and solutions to that problem are hacky

devchat.go
```diff
@@ -437,12 +438,51 @@ func (u *user) pickUsername(possibleName string) {
...
+			return colorArr[i]
+		}
+	}
+	return colorArr[0]
+}
+
+func (u *user) rainbowColor() {
+	var rainbow = []color.Color{
+		*color.New(color.FgHiRed),
+		*color.New(color.FgHiGreen),
```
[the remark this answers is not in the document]
ciearius: Correct me if I am wrong, but didn't you point out earlier that I shouldn't use references to those variables at the top? Although I can't find that comment anymore...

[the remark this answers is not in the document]
quackduck: Ah no that was to say that you shouldn’t take the pointer of variables that are already pointers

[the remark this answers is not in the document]
quackduck: Here I mean to say to use the global variables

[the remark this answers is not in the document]
quackduck: For example, `*color.New(color.FgHiRed)` can entirely be replaced with the global variable `red`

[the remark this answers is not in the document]
quackduck: Get what I mean?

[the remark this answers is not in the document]
quackduck: And by using the global variables we get more readable, succinct code and also prevent unnecessary allocations

ciearius:
I've rewritten parts of it now... We are now using "colorstyles"

IamTregsthedev:
Make this for Pride Month, It would be so nice like to me and everyone. Don't make it serect

quackduck:
Good idea about pride month

devchat.go
```diff
@@ -89,6 +137,12 @@ var (
 		"1eab2de20e41abed903ab2f22e7ff56dc059666dbe2ebbce07a8afeece8d0424"} // Shok
 )
 
+func buildStyleApplicator(c color.Color) styleApplication {
```
quackduck, line 140: Make this take a pointer!

devchat.go
```diff
@@ -1171,3 +1245,10 @@ func remove(s []*user, a *user) []*user {
 	}
 	return s
 }
+
+type styleApplication func(string) string
```
quackduck, line 1249: Nah you don’t really need this new type, let the definition of color styles directly use a function

devchat.go
quackduck: Also might be a good idea to have a shorter name like mkStyle() but I can do that myself too

devchat.go
```diff
@@ -45,19 +46,66 @@ var (
...
+		{"green", buildStyleApplicator(*green)},
+		{"cyan", buildStyleApplicator(*cyan)},
+		{"magenta", buildStyleApplicator(*magenta)},
+		{"yellow", buildStyleApplicator(*yellow)},
+		{"blue", buildStyleApplicator(*blue)},
+		{"black", buildStyleApplicator(*black)},
+		{"easter", buildStyleApplicator(*color.New(color.BgMagenta, color.FgHiYellow))},
+		{"baby", buildStyleApplicator(*color.New(color.BgBlue, color.FgHiMagenta))},
+		{"hacker", buildStyleApplicator(*color.New(color.FgHiGreen, color.BgBlack))},
+		//TODO: find a way to support special colors which rely on the user object
```
quackduck, line 72: Great point! Thought about this a bit and I realized that we can simply have the applicator print the username directly in the color to be added without first stripping its own colors. Then we could remove the extra ANSI reset code at the end.

devchat.go
```diff
@@ -45,19 +46,66 @@ var (
...
+	cyan        = color.New(color.FgHiCyan)
+	magenta     = color.New(color.FgHiMagenta)
+	yellow      = color.New(color.FgHiYellow)
+	blue        = color.New(color.FgHiBlue)
+	black       = color.New(color.FgBlack)
+	white       = color.New(color.FgHiWhite)
 	devbot      = "" // initialized in main
 	startupTime = time.Now()
 
+	colorstyles = []colorstyle{
```
quackduck, line 60: Nicer to make this an `[]*colorstyle` instead.

devchat.go
```diff
@@ -45,19 +46,66 @@ var (
...
+					blue,
+					cyan,
+					yellow,
+					green,
+					white,
+				}
+
+				var stripped = stripansi.Strip(a)
+				var buf = ""
+				colorOffset := rand.Intn(len(rainbow) - 1)
```
quackduck, line 97:
```suggestion
				colorOffset := rand.Intn(len(rainbow))
```
rand.Intn always returns a number less than the argument (`[0,n)`)

devchat.go
```diff
@@ -89,6 +137,12 @@ var (
 		"1eab2de20e41abed903ab2f22e7ff56dc059666dbe2ebbce07a8afeece8d0424"} // Shok
 )
 
+func buildStyleApplicator(c color.Color) styleApplication {
+	return func(s string) string {
+		return c.Sprint(stripansi.Strip(s))
+	}
+}
```
quackduck, line 144:
```suggestion
func buildStyleApplicator(c *color.Color) *styleApplication {
	return &(func(s string) string {
		return c.Sprint(stripansi.Strip(s))
	})
}
```

devchat.go
```diff
@@ -437,19 +491,37 @@ func (u *user) pickUsername(possibleName string) {
...
+func (u *user) changeColor(colorName string) {
+	color := getColor(colorName)
+	u.changeColorstyle(color)
 }
 
-func (u *user) changeColor(color color.Color) {
-	u.name = color.Sprint(stripansi.Strip(u.name))
-	u.color = color
+// Applies color from colorstyle
+func (u *user) changeColorstyle(color colorstyle) {
```
quackduck, line 505: Merge this function into `changeColor` directly (we're trying to make the entire outward API work with colors as named strings)

devchat.go
```diff
@@ -1171,3 +1245,10 @@ func remove(s []*user, a *user) []*user {
 	}
 	return s
 }
+
+type styleApplication func(string) string
+
+type colorstyle struct {
+	name  string
+	apply styleApplication
+}
```
quackduck, line 1254:
```suggestion
type colorstyle struct {
	name  string
	apply func(string) string
}
```

devchat.go
```diff
@@ -437,19 +491,37 @@ func (u *user) pickUsername(possibleName string) {
...
 
+// Turns name into an applyable color (defaults to white)
+func getColor(name string) colorstyle {
+	for i := 0; i < len(colorstyles); i++ {
+		if colorstyles[i].name == name {
+			return colorstyles[i]
+		}
+	}
+	return colorstyles[0]
+}
```
quackduck, line 523:
```suggestion
func getColor(name string) *colorstyle {
	for i := range colorstyles {
		if colorstyles[i].name == name {
			return colorstyles[i]
		}
	}
	return colorstyles[0]
}
```

devchat.go
quackduck:
```suggestion
	colorstyles = []*colorstyle{
```

devchat.go
```diff
@@ -45,19 +46,66 @@ var (
...
+		{"red", buildStyleApplicator(*red)},
+		{"green", buildStyleApplicator(*green)},
+		{"cyan", buildStyleApplicator(*cyan)},
+		{"magenta", buildStyleApplicator(*magenta)},
+		{"yellow", buildStyleApplicator(*yellow)},
+		{"blue", buildStyleApplicator(*blue)},
+		{"black", buildStyleApplicator(*black)},
+		{"easter", buildStyleApplicator(*color.New(color.BgMagenta, color.FgHiYellow))},
+		{"baby", buildStyleApplicator(*color.New(color.BgBlue, color.FgHiMagenta))},
+		{"hacker", buildStyleApplicator(*color.New(color.FgHiGreen, color.BgBlack))},
```
quackduck, line 71:
```suggestion
		{"white", buildStyleApplicator(white)},
		{"red", buildStyleApplicator(red)},
		{"green", buildStyleApplicator(green)},
		{"cyan", buildStyleApplicator(cyan)},
		{"magenta", buildStyleApplicator(magenta)},
		{"yellow", buildStyleApplicator(yellow)},
		{"blue", buildStyleApplicator(blue)},
		{"black", buildStyleApplicator(black)},
		{"easter", buildStyleApplicator(color.New(color.BgMagenta, color.FgHiYellow))},
		{"baby", buildStyleApplicator(color.New(color.BgBlue, color.FgHiMagenta))},
		{"hacker", buildStyleApplicator(color.New(color.FgHiGreen, color.BgBlack))},
```

quackduck:
Okay so an unintended side effect is that secret colors are no longer secret. We could fix this by making a separate array for secret colors that `getColor()` searches for after searching in the main array `colorstyles`. This way, random picks will only pick from the main array, but secret colors still work.

devchat.go
quackduck: For some reason I'm not able to make this a nice little clickable suggestion

quackduck:
I wonder if this compiles 😆

devchat.go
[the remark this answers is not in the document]
quackduck: Done! I committed this change.

quackduck:
If you'd like it @ciearius I could simply accept this and make the changes needed myself. 

Or would you like to make the changes?

IamTregsthedev:
I can try, tell me how to fix it?

quackduck:
Just check the comments above, I don't think you can directly make changes tho

quackduck:
We'll have to wait for @ciearius to tell us what he'd like more :)

ciearius:
My  understanding of golang is quite limited. I am having trouble understanding why you would use pointers there and in other places you don't? I'd recommend you go ahead and clone this fork, make your changes and then we can look at it again! Right now it doesn't compile... :cry:

quackduck:
Wait where am I not using pointers, I wanna use em everywhere

quackduck:
Also, could you please post the compile error message here @ciearius, thanks!

ciearius:
Which one? There are plenty 😅

quackduck:
Send all of them lol @ciearius

ciearius:
Here you are:
```
./devchat.go:84:4: cannot use func literal (type func(string) string) as type *func(string) string in field value
./devchat.go:142:9: cannot take the address of func literal
./devchat.go:504:20: cannot use color (type *colorstyle) as type colorstyle in argument to u.changeColorstyle
./devchat.go:510:22: cannot call non-function color.apply (type *func(string) string)
./devchat.go:1214:96: cannot call non-function colorstyles[int(i) % len(colorstyles)].apply (type *func(string) string)
```

quackduck:
fixed them all

quackduck:
Turns out functions as variables are already pointers

quackduck:
Okay we are ready to accept! Thank you so so so much @ciearius for contributing to Devzat! 💖

PR #26, opened 2021-06-09
Hide hangman word otherwise kinda pointless.

TommyPujol06:
When starting a hangman game, the word is shown, which makes it pointless since everyone can cheat. To avoid this, the word is read as if it were a password and only the person starting the game knows it.

### DIFF
--- a/devchat.go
+++ b/devchat.go
@@ -109,6 +109,7 @@ var (
 	tttGame       = new(ttt.Board)
 	currentPlayer = ttt.X
 	hangGame      = new(hangman)
+	hangGameStarted = false
 
 	admins = []string{"d84447e08901391eb36aa8e6d9372b548af55bee3799cd3abb6cdd503fdf2d82", // Ishan Goel
 		"f5c7f9826b6e143f6e9c3920767680f503f259570f121138b2465bb2b052a85d", // Ella Xu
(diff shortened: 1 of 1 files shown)


quackduck:
Actually the word isn't shown!

devchat.go
```diff
@@ -591,27 +592,32 @@ func runCommands(line string, u *user, isSlack bool) {
 		return
 	}
 	if strings.HasPrefix(line, "/hang") {
-		rest := strings.TrimSpace(strings.TrimPrefix(line, "/hang"))
-		if len(rest) > 1 {
-			u.writeln(u.name, line)
```
quackduck, line 596: see this line? it makes sure only the sending user sees their command line

TommyPujol06: Yeah didn't notice.

devchat.go
```diff
@@ -591,27 +592,32 @@ func runCommands(line string, u *user, isSlack bool) {
...
-			hangGame = &hangman{rest, 15, " "} // default value of guesses so empty space is given away
+		if !hangGameStarted {
+			word, err := u.term.ReadPassword("Word (Won't be displayed): ")
```
quackduck, line 596: Using ReadPassword will turn off echo, which can be confusing to people not so used to the command line.

devchat.go
quackduck: It's okay!

quackduck:
It's okay! I can see how the UI right now could make a user think that their word is displayed (would love suggestions on how that could be mitigated). Anyways, thanks for helping out Devzat!

devchat.go
TommyPujol06: Yes I'm aware of that, it is the reason I added `(Won't be displayed)`. But doesn't matter any more since this change is not needed.

PR #28, opened 2021-06-10
Add ./name as an alternative to ./nick

arashnrim:
I think the title of this PR is explanatory enough. Not entirely certain if it'll work, but I'm guessing it should!


quackduck:
oop, closed by mistake, it'll work, thanks for your contribution to Devzat 💖

PR #29, opened 2021-06-14
Basic reply command.

TAG-Epic:
Code might be buggy, works for me atleast

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -75,8 +75,8 @@ func runCommands(line string, u *user, isSlack bool) {
 			return
 		}
 		u.lastDM.lastDM = u
-		u.lastDM.writeln(u.name+" <- ", message)
-		u.writeln(u.name+" -> ", message)
+		u.lastDM.writeln(u.name+" -> ", message)
+		u.writeln(user.lastDM.name+" <- ", message)
 		return
 	}
 	if strings.HasPrefix(line, "/hang") {
--- a/commands.go
+++ b/commands.go
@@ -2,13 +2,14 @@ package main
 
 import (
 	"fmt"
-	"github.com/acarl005/stripansi"
-	"github.com/shurcooL/tictactoe"
 	"sort"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/acarl005/stripansi"
+	"github.com/shurcooL/tictactoe"
 )
 
 // runCommands parses a line of raw input from a user and sends a message as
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -56,10 +57,28 @@ func runCommands(line string, u *user, isSlack bool) {
...
+		if len(message) == 0 {
+			u.writeln("devbot", "You need to add a message.")
+			return
+		}
+		if u.lastDM == nil {
+			u.writeln("devbot", "There is no one to reply to.")
+			return
+		}
+		u.lastDM.lastDM = u
+		u.lastDM.writeln(u.name+" <- ", message)
```
quackduck, line 78: The arrow should be the other way round

commands.go
```diff
@@ -56,10 +57,28 @@ func runCommands(line string, u *user, isSlack bool) {
...
+		u.writeln(u.name+" -> ", message)
```
quackduck, line 79: Arrow should be other way around and should be used with user.lastDm

quackduck:
I think this will work

quackduck:
However, why do this? I would be able to see the rationale behind a feature that makes it so without using any command, all next messages go as DMs to lastDm, but not for a command that you still have to have at the front of each message anyway.

TAG-Epic:
> However, why do this? I would be able to see the rationale behind a feature that makes it so without using any command, all next messages go as DMs to lastDm, but not for a command that you still have to have at the front of each message anyway.

People have hard to spell names mostly

quackduck:
You can simply use the up arrow

commands.go
```diff
@@ -56,10 +57,28 @@ func runCommands(line string, u *user, isSlack bool) {
...
+		message := strings.TrimSpace(strings.TrimPrefix(line, "./reply"))
+		if len(message) == 0 {
+			u.writeln("devbot", "You need to add a message.")
+			return
+		}
+		if u.lastDM == nil {
+			u.writeln("devbot", "There is no one to reply to.")
+			return
+		}
+		u.lastDM.lastDM = u
```
quackduck, line 77: I love this line haha

TAG-Epic:
> You can simply use the up arrow

Mobile users?

GitOpsBot (commented):
# GoGitOps Review

## Grade: A+ (93.3%)

### Files: 7

### Issues: 3

### gofmt: 100%

### go_vet: 100%

### gocyclo: 57%

	util.go
	colors.go
	commands.go
		Line 19: warning: cyclomatic complexity 89 of function runCommands() is high (> 15) (gocyclo)
### golint: 85%

	colors.go
### license: 100%

### ineffassign: 100%

### misspell: 100%




quackduck:
> Mobile users?
smol segment of users, plus you still need to do the typing, plus mobile users can simply copy the string they need each time

quackduck:
Make ./reply either take "on" or "off" and then make it so after being switched on, further messages are automatically turned into DMS to u.LastDm, and off makes messages global again.

I think that's much more useful

quackduck:
Let me know if in that case there's a better name we could give to this command, or whether we should add it into ./dm:
`./dm replies on` (we'll need to make sure there's never a user called replies, see u.pickName)

GitOpsBot (changes requested):
# GoGitOps Review

## Grade: A (88.2%)

### Files: 7

### Issues: 3

### gofmt: 85%

```diff
- This PR will be rejected until gofmt is 100%
```
	commands.go
### go_vet: 100%

### gocyclo: 57%

	commands.go
		Line 19: warning: cyclomatic complexity 90 of function runCommands() is high (> 15) (gocyclo)
	util.go
	colors.go
### golint: 85%

	colors.go
### license: 100%

### ineffassign: 100%

### misspell: 100%




TAG-Epic:
Remaking this on another branch.

PR #32, opened 2021-06-14
Toggling only talking to private messages

TAG-Epic:
Continutation from #29. basically just lets you private message someone by doing ./room @<mention> and ./room to leave to the previous room

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -213,7 +213,7 @@ func runCommands(line string, u *user, isSlack bool) {
 	if strings.HasPrefix(line, "./room") && !isSlack {
 		if u.messaging != nil {
 			u.messaging = nil
-			u.writeln(devbot, fmt.Sprintf("Left private chat and put you back in %s", u.room.name))
+			u.writeln(devbot, "Left private chat")
 			return
 		}
 		rest := strings.TrimSpace(strings.TrimPrefix(line, "./room"))
--- a/commands.go
+++ b/commands.go
@@ -20,6 +20,11 @@ func runCommands(line string, u *user, isSlack bool) {
 	if line == "" {
 		return
 	}
+	if u.messaging != nil && !(strings.HasPrefix(line, "./") || strings.HasPrefix(line, "=")) {
+		u.writeln(u.messaging.name+" <- ", line)
+		u.messaging.writeln(u.name+" -> ", line)
+		return
+	}
 
 	sendToSlack := true
 	b := func(senderName, msg string) {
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -224,15 +234,27 @@ func runCommands(line string, u *user, isSlack bool) {
...
+			restSplit := strings.Fields(strings.TrimPrefix(rest, "@"))
+			peer, ok := findUserByName(u.room, restSplit[0])
+			if !ok {
+				u.writeln(devbot, "No such person lol, who you wanna dm? (you might be in the wrong room)")
+				return
+			}
+			u.messaging = peer
+			u.writeln(devbot, fmt.Sprintf("Now messaging %s. To leave say \n>./room", peer.name))
+			return
+		} else {
```
quackduck, line 255: Don't have an else statement since this will catch on to any message if I understand correctly

quackduck: Nope, I did not understand correctly

commands.go
```diff
@@ -206,6 +211,11 @@ func runCommands(line string, u *user, isSlack bool) {
 		return
 	}
 	if strings.HasPrefix(line, "./room") && !isSlack {
+		if u.messaging != nil {
+			u.messaging = nil
+			u.writeln(devbot, fmt.Sprintf("Left private chat and put you back in %s", u.room.name))
```
quackduck, line 216: Make this only say "left private chat":
```suggestion
			u.writeln(devbot, "Left private chat")
```
So that if a user is further deciding to join another room or dm someone else it won't say inconsistent stuff.

commands.go
```diff
@@ -20,6 +20,11 @@ func runCommands(line string, u *user, isSlack bool) {
 	if line == "" {
 		return
 	}
+	if u.messaging != nil && !(strings.HasPrefix(line, "./") || strings.HasPrefix(line, "=")) {
```
quackduck, line 23: Side effect of having `strings.HasPrefix(line, "./")` in this condition is that if I send `./help` while in a private chat dm, it will be broadcast to a whole room. Fix by just removing that part of the condition:
```suggestion
	if u.messaging != nil && !strings.HasPrefix(line, "=") {
```
Alternatively, move this block after the check for whether the line is a DM (better design IMO)

commands.go
GitOpsBot, line 255: warning: if block ends with a return statement, so drop this else and outdent its block (golint)

GitOpsBot (commented):
# GoGitOps Review

## Grade: A+ (91.6%)

### Files: 7

### Issues: 3

### gofmt: 100%

### go_vet: 100%

### gocyclo: 57%

	commands.go
		Line 19: warning: cyclomatic complexity 93 of function runCommands() is high (> 15) (gocyclo)
	util.go
	colors.go
### golint: 71%

	colors.go
	commands.go
		Line 255: warning: if block ends with a return statement, so drop this else and outdent its block (golint)
### license: 100%

### ineffassign: 100%

### misspell: 100%




commands.go
quackduck: ooohhhh, smart bot

quackduck:
Thanks so much for contributing _yet again_ to Devzat! 💖

PR #33, opened 2021-06-14
Anti-Spambot

TAG-Epic:
A small anti-spam system which triggers when someone sends 10 messages in one second. Doubt it would take down ./color random spamming but if it does its just to tweak the max messages per second value.

Just to take down spambots if someone decides to spam it

### DIFF
--- a/devchat.go
+++ b/devchat.go
@@ -5,9 +5,6 @@ import (
 	_ "embed"
 	"encoding/hex"
 	"fmt"
-	"github.com/acarl005/stripansi"
-	"github.com/gliderlabs/ssh"
-	terminal "golang.org/x/term"
 	"io"
 	"log"
 	"math"
@@ -20,6 +17,10 @@ import (
 	"sync"
 	"syscall"
 	"time"
+
+	"github.com/acarl005/stripansi"
+	"github.com/gliderlabs/ssh"
+	terminal "golang.org/x/term"
 )
 
 var (
@@ -45,6 +46,8 @@ var (
 	bansMutex       = sync.Mutex{}
 	idsInMinToTimes = make(map[string]int, 10)
 	idsInMinMutex   = sync.Mutex{}
+
+	antispamMessages = make(map[string]int)
 )
 
 type room struct {
@@ -92,7 +95,12 @@ func main() {
 			"If you still can't join, try joining back in 2 minutes. If you _still_ can't join, make an issue at github.com/quackduck/devzat/issues", true)
 		os.Exit(0)
 	}()
-
+	go func() {
+		for {
+			time.Sleep(1 * time.Second)
+			antispamMessages = make(map[string]int)
+		}
+	}()
 	ssh.Handle(func(s ssh.Session) {
 		u := newUser(s)
 		if u == nil {
@@ -337,6 +345,27 @@ func (u *user) repl() {
 			continue
 		}
 		u.term.Write([]byte(strings.Repeat("\033[A\033[2K", int(math.Ceil(float64(len([]rune(u.name+line))+2)/(float64(u.win.Width))))))) // basically, ceil(length of line divided by term width)
+
+		// Anti spam check
+		currentMessages, exists := antispamMessages[u.id]
+		if !exists {
+			currentMessages = 0
+		}
+		currentMessages += 1
+		antispamMessages[u.id] = currentMessages
+		if currentMessages >= 10 {
+			bans = append(bans, u.addr)
+			u.writeln(devbot, "Anti-Spam triggered")
+			u.close(red.Paint(u.name + " has been banned."))
+			for _, room := range rooms {
+				for _, user := range room.users {
+					if user.id == u.id {
+						user.writeln(devbot, "Anti-Spam triggered on another account.")
+						u.close(red.Paint(u.name + " has been banned."))
+					}
+				}
+			}
+		}
 
 		runCommands(line, u, false)
 	}
--- a/devchat.go
+++ b/devchat.go
@@ -48,6 +48,7 @@ var (
 	idsInMinMutex   = sync.Mutex{}
 
 	antispamMessages = make(map[string]int)
+	antispamMutex    = sync.Mutex{}
 )
 
 type room struct {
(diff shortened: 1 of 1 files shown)


devchat.go
```diff
@@ -338,6 +346,27 @@ func (u *user) repl() {
 		}
 		u.term.Write([]byte(strings.Repeat("\033[A\033[2K", int(math.Ceil(float64(len([]rune(u.name+line))+2)/(float64(u.win.Width))))))) // basically, ceil(length of line divided by term width)
 
+		// Anti spam check
+		currentMessages, exists := antispamMessages[u.id]
+		if !exists {
+			currentMessages = 0
+		}
+		currentMessages += 1
+		antispamMessages[u.id] = currentMessages
```
quackduck, line 355: Since maps give zero values when given a missing key, 349 - 355 can be shortened to antispamMessages[u.id]++

devchat.go
```diff
@@ -338,6 +346,27 @@ func (u *user) repl() {
...
+		if currentMessages >= 10 {
+			bans = append(bans, u.addr)
+			u.writeln(devbot, "Anti-Spam triggered")
+			u.close(red.Paint(u.name + " has been banned."))
+			for _, room := range rooms {
```
quackduck, line 360: This for loop isn’t necessary, close already broadcasts it

devchat.go
```diff
@@ -338,6 +346,27 @@ func (u *user) repl() {
...
+		currentMessages, exists := antispamMessages[u.id]
+		if !exists {
+			currentMessages = 0
+		}
+		currentMessages += 1
+		antispamMessages[u.id] = currentMessages
+		if currentMessages >= 10 {
+			bans = append(bans, u.addr)
+			u.writeln(devbot, "Anti-Spam triggered")
+			u.close(red.Paint(u.name + " has been banned."))
```
quackduck, line 359: Maybe add that they were banned due to spam :)

quackduck:
Should we kick instead of ban? What do you think? (Great PR idea btw)

GitOpsBot (commented):
# GoGitOps Review

## Grade: A+ (93.3%)

### Files: 7

### Issues: 3

### gofmt: 100%

### go_vet: 100%

### golint: 85%

	colors.go
### gocyclo: 57%

	commands.go
		Line 19: warning: cyclomatic complexity 87 of function runCommands() is high (> 15) (gocyclo)
	util.go
	colors.go
### ineffassign: 100%

### license: 100%

### misspell: 100%




quackduck:
Also, could you make both the time and the messages numbers larger (for example, suppose I copy a tiny section of a wiki page and paste it in, I'll get banned)

I think 50 messages + 5 seconds is a nicer limit (I can change that, no worries)

devchat.go
```diff
@@ -92,7 +95,12 @@ func main() {
 			"If you still can't join, try joining back in 2 minutes. If you _still_ can't join, make an issue at github.com/quackduck/devzat/issues", true)
 		os.Exit(0)
 	}()
-
+	go func() {
```
quackduck, line 98: Using this approach (refreshing the entire map each second) can lead to an unexpected side effect:

Suppose someone posts 10 messages in 50 seconds half a second before the map is due to be refreshed, then, even tho they _should_ be banned, they won't since the map gets refreshed mid-way. To fix this, edit the part where the message counter is incremented and make it decrease that ID's count after one second. (use `time.AfterFunc`)

devchat.go
quackduck, line 355:
```suggestion
		antispamMessages[u.id]++
		time.AfterFunc(1*time.Second, func() {
			antispamMessages[u.id]--
		}
```

devchat.go
```diff
@@ -92,7 +95,12 @@ func main() {
...
+		for {
+			time.Sleep(1 * time.Second)
+			antispamMessages = make(map[string]int)
+		}
+	}()
```
quackduck, line 103:
```suggestion
```

devchat.go
quackduck:
```suggestion
			u.close(red.Paint(u.name + " has been banned for spamming"))
```

devchat.go
```diff
@@ -338,6 +346,27 @@ func (u *user) repl() {
...
+			for _, room := range rooms {
+				for _, user := range room.users {
+					if user.id == u.id {
+						user.writeln(devbot, "Anti-Spam triggered on another account.")
+						u.close(red.Paint(u.name + " has been banned."))
+					}
+				}
+			}
+		}
+
```
quackduck, line 369:
```suggestion
```

devchat.go
```diff
@@ -45,6 +46,8 @@ var (
 	bansMutex       = sync.Mutex{}
 	idsInMinToTimes = make(map[string]int, 10)
 	idsInMinMutex   = sync.Mutex{}
+
+	antispamMessages = make(map[string]int)
```
quackduck, line 50:
```suggestion
	antispamMessages = make(map[string]int)
	antispamMutex    = sync.Mutex{}
```

devchat.go
quackduck: Since maps give zero values when given a missing key, 349 - 355 can be shortened to antispamMessages[u.id]++

devchat.go
quackduck: This for loop isn’t necessary, close already broadcasts it

devchat.go
```diff
@@ -338,6 +346,27 @@ func (u *user) repl() {
...
 		u.term.Write([]byte(strings.Repeat("\033[A\033[2K", int(math.Ceil(float64(len([]rune(u.name+line))+2)/(float64(u.win.Width))))))) // basically, ceil(length of line divided by term width)
 
+		// Anti spam check
+		currentMessages, exists := antispamMessages[u.id]
+		if !exists {
+			currentMessages = 0
+		}
+		currentMessages += 1
+		antispamMessages[u.id] = currentMessages
+		if currentMessages >= 10 {
```
quackduck, line 356:
```suggestion
		if currentMessages >= 50 {
```

quackduck:
Thank you so much for implementing this! 💖

PR #40, opened 2021-06-16
added https:// to twitter url, makes it aktually openable from terminal

patrickhener:
So this way it makes the link openable from terminal. Let's say you are using `gnome-terminal` like I do on linux, and you right click that link you will get a `open in browser`. This way you can actually directly open that link without needing to copy paste it.

### DIFF
--- a/twitter.go
+++ b/twitter.go
@@ -57,7 +57,7 @@ func sendCurrentUsersTwitterMessage() {
 			mainRoom.broadcast(devbot, "err: "+err.Error(), true)
 			return
 		}
-		mainRoom.broadcast(devbot, "https://twitter.com/"+t.User.ScreenName+"/status/"+t.IDStr, true)
+		mainRoom.broadcast(devbot, "https\://twitter.com/"+t.User.ScreenName+"/status/"+t.IDStr, true)
 	}()
 	//broadcast(devbot, tweet.Entities.Urls)
 }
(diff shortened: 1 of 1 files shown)


GitOpsBot (commented):
# GoGitOps Review

## Grade: A (88.2%)

### Files: 7

### Issues: 3

### gofmt: 85%

```diff
- A new PR should be opened to fix trivial gofmt issues
```
	util.go
### go_vet: 100%

### gocyclo: 57%

	commands.go
		Line 19: warning: cyclomatic complexity 93 of function runCommands() is high (> 15) (gocyclo)
	util.go
		Line 176: warning: cyclomatic complexity 30 of function devbotChat() is high (> 15) (gocyclo)
	colors.go
		Line 106: warning: cyclomatic complexity 19 of function getStyle() is high (> 15) (gocyclo)
### golint: 85%

	util.go
		Line 267: warning: should omit 2nd value from range; this loop is equivalent to `for i := range ...` (golint)
### license: 100%

### ineffassign: 100%

### misspell: 100%




quackduck:
Actually I’ve made it not have https on purpose because when the markdown lib renders urls it expands them to the markdown syntax, maybe escaping the h in https can stop that.

patrickhener:
Oh I see. Or we could see if you can stop the markdown library from doing that? Cause if you paste a link it will also render it strange. I see no benefit in converting the links into markdown style in the chat app?

quackduck:
yah I'll fix that

quackduck:
Thanks for contributing!

PR #45, opened 2021-06-29
Fixed deletion of messages on client.

Arkaeriit:
Sometime, especialy with small terminals
and fancy name colors, the last printed
messages was deleted because the colored
char were counted in the printed message
size. This is now fixed.

### DIFF
--- a/devchat.go
+++ b/devchat.go
@@ -51,7 +51,6 @@ type room struct {
 
 type user struct {
 	name          string
-	raw_name      string
 	session       ssh.Session
 	term          *terminal.Terminal
 	bell          bool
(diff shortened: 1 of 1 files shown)


quackduck:
Actually, this won’t work when names are ./nick-ed or when they need to be renamed. Instead, just use the stripansi.Strip function where the length is passed on to the markdown render function in u.writeln

quackduck:
I would never have guessed that _this_ was the bug, you’re really great at spotting bugs!

quackduck:
This is absolutely perfect, you make great, easy to work with PRs, thank you _so_ much!

quackduck:
> My fix works when you nick yourself as the nick command calls the pickUsername functions which is where the variable 'raw_string' is eddied.

Oh right

PR #47, opened 2021-07-09
24h format

Arkaeriit:
Added option to display time in 24h format.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -33,7 +33,7 @@ var (
 		{"hang", hangCMD, "<char|word>", "Play hangman"}, // won't actually run, here just to show in docs
 		{"tic", ticCMD, "<cell num>", "Play tic tac toe!"},
 		{"cd", cdCMD, "#room/user", "Join #room, DM user or run cd to see a list"}, // won't actually run, here just to show in docs
-		{"tz", tzCMD, "<zone> [24h]", "Set your IANA timezone (like tz Asia/Dubai).  Optional argument for 24h format."},
+		{"tz", tzCMD, "<zone> [24h]", "Set your IANA timezone (like tz Asia/Dubai) and optionally set 24h"},
 		{"nick", nickCMD, "<name>", "Change your username"},
 		{"rest", commandsRestCMD, "", "Uncommon commands list"}}
 	cmdsRest = []cmd{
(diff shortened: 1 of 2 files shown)


devchat.go
```diff
@@ -195,6 +196,8 @@ func newUser(s ssh.Session) *user {
 		win:           w,
 		lastTimestamp: time.Now(),
 		joinTime:      time.Now(),
+		timezone:      nil,
```
quackduck, line 199: nil is the zero value so this line can be deleted

commands.go
```diff
@@ -33,7 +33,7 @@ var (
 		{"hang", hangCMD, "<char|word>", "Play hangman"}, // won't actually run, here just to show in docs
 		{"tic", ticCMD, "<cell num>", "Play tic tac toe!"},
 		{"cd", cdCMD, "#room/user", "Join #room, DM user or run cd to see a list"}, // won't actually run, here just to show in docs
-		{"tz", tzCMD, "<zone>", "Set your IANA timezone (like tz Asia/Dubai)"},
+		{"tz", tzCMD, "<zone> [format]", "Set your IANA timezone (like tz Asia/Dubai). Set format to 24 to have hour in the 0-23 range."},
```
quackduck, line 36: Change format to 24 and shorten that help message to “set your IANA ... and set 24h clock

Arkaeriit: Good idea, I was wondering how I could shorten the help message.

devchat.go
Arkaeriit: Isn't it better to clearly state that having the timezone set to nil is an expected behavior and the default value?
Otherwise, someone might thought that letting that out is a bug.

commands.go
```diff
@@ -33,7 +33,7 @@ var (
 		{"hang", hangCMD, "<char|word>", "Play hangman"}, // won't actually run, here just to show in docs
 		{"tic", ticCMD, "<cell num>", "Play tic tac toe!"},
 		{"cd", cdCMD, "#room/user", "Join #room, DM user or run cd to see a list"}, // won't actually run, here just to show in docs
-		{"tz", tzCMD, "<zone>", "Set your IANA timezone (like tz Asia/Dubai)"},
+		{"tz", tzCMD, "<zone> [24h]", "Set your IANA timezone (like tz Asia/Dubai).  Optional argument for 24h format."},
```
quackduck, line 36:
```suggestion
		{"tz", tzCMD, "<zone> [24h]", "Set your IANA timezone (like tz Asia/Dubai) and optionally set 24h"},
```

quackduck:
Perfect, thanks

devchat.go
quackduck: Hmm, I already leave out all zero values

PR #48, opened 2021-07-09
Documented loggin with nickname in README

Arkaeriit:
Some users might be unhappy about logging in and having their real name showed to everyone.
To prevent this, I told in the readme how to log in with a nickname.


quackduck:
Thanks!

PR #52, opened 2021-07-29
Fixed bugs with the bell.

Arkaeriit:
Sometime, users were ping while
their bell was off or for every
messages while their bell were
on default mode. This is fixed.


quackduck:
Oh, perfect, thanks so much

PR #55, opened 2021-07-29
Generate the user ID based on their public key instead of IP

TommyPujol06:
Since IP's can change, it's better if we get the ID from the public key. This has a few benefits, such as:

* People in the same network/household will be able to connect to the chat even if one person is banned.
* Admin takeover is much harder since instead of needing to be in the same network, you will need their public/private key pair.
* A person can connect from anywhere and as long as they have their key they can prove their legitimacy. (i.e. a random person called "ishan" on the chat won't be able to act the real ishan since your ID will be static and thus the fake ishan will be detected).

In the case that a user logged in without using their key pair, we will still get their ID from the IP, but we still gain all the benefits of pubkey authentication.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -339,8 +339,7 @@ func idCMD(line string, u *user, _ bool) {
 		u.room.broadcast("", "User not found")
 		return
 	}
-	u.room.broadcast("IP based id", victim.uid)
-	u.room.broadcast("Pubkey based id", victim.cid)
+	u.room.broadcast("", victim.id)
 }
 
 func nickCMD(line string, u *user, _ bool) {
--- a/commands.go
+++ b/commands.go
@@ -38,7 +38,7 @@ var (
 		{"rest", commandsRestCMD, "", "Uncommon commands list"}}
 	cmdsRest = []cmd{
 		{"people", peopleCMD, "", "See info about nice people who joined"},
-		{"id", idCMD, "<user>", "Get a unique ID for a user (hashed IP)"},
+		{"id", idCMD, "<user>", "Get a unique ID for a user (hashed IP | SSH Public key)"},
 		{"eg-code", exampleCodeCMD, "", "Example syntax-highlighted code"},
 		{"banIP", banIPCMD, "<IP>", "Ban an IP (admin)"},
 		{"ban", banCMD, "<user>", "Ban <user> (admin)"},
@@ -339,7 +339,8 @@ func idCMD(line string, u *user, _ bool) {
 		u.room.broadcast("", "User not found")
 		return
 	}
-	u.room.broadcast("", victim.id)
+	u.room.broadcast("IP based id", victim.uid)
+	u.room.broadcast("Pubkey based id", victim.cid)
 }
 
 func nickCMD(line string, u *user, _ bool) {
--- a/commands.go
+++ b/commands.go
@@ -38,7 +38,7 @@ var (
 		{"rest", commandsRestCMD, "", "Uncommon commands list"}}
 	cmdsRest = []cmd{
 		{"people", peopleCMD, "", "See info about nice people who joined"},
-		{"id", idCMD, "<user>", "Get a unique ID for a user (hashed IP | SSH Public key)"},
+		{"id", idCMD, "<user>", "Get a unique ID for a user (hashed key)"},
 		{"eg-code", exampleCodeCMD, "", "Example syntax-highlighted code"},
 		{"banIP", banIPCMD, "<IP>", "Ban an IP (admin)"},
 		{"ban", banCMD, "<user>", "Ban <user> (admin)"},
--- a/devchat.go
+++ b/devchat.go
@@ -186,7 +186,6 @@ func newUser(s ssh.Session) *user {
 	hash := sha256.New()
 	pubkey := s.PublicKey()
 	if pubkey != nil {
-		fmt.Printf("key: %s\n", pubkey.Marshal())
 		hash.Write([]byte(pubkey.Marshal()))
 	} else { // If we can't get the public key fall back to the IP.
 		hash.Write([]byte(host))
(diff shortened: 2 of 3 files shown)


quackduck:
1. People can easily un-ban themselves (this is not too bad, since this still roots out automated spammers)
2. Existing ID data will need to be changed (this used to be a huge issue cause of the `all` command, but now it is gone. Yet, we'll need to change the admin data and the ban data)

quackduck:
this used to be how IDs used to work, but i changed it as soon as I realized people could easily un-ban themselves.

TommyPujol06:
OK, good point about the bans and admins, I think we could have two IDs, one based off the IP and the other based on the private key. We can use the public key ID for having configurations (what I said on discord about having a small database to keep track of user preferences) and the IP ID for bans, etc. 

Eventually, I think it'd be in the best interest to move admin ID's to be public key based, since admins will be able to use their privileges regardless of where they are logging in from.  It also adds a layer of security for admins since it will be harder to get their permissions. 

As of right now, I will leave the admin checking to be IP based, however I think it would be good to change it in the near future even if it means that we will need to manually re-add all admins.

GitOpsBot (commented):
# GoGitOps Review

## Grade: A+ (96.6%)

### Files: 7

### Issues: 2

### gofmt: 100%

### go_vet: 100%

### golint: 100%

### gocyclo: 71%

	util.go
		Line 171: warning: cyclomatic complexity 29 of function devbotChat() is high (> 15) (gocyclo)
	colors.go
		Line 98: warning: cyclomatic complexity 19 of function getStyle() is high (> 15) (gocyclo)
### ineffassign: 100%

### license: 100%

### misspell: 100%




quackduck:
I think it's ok to have only pubkey IDs, since I'm sure people who do get banned won't really read the source code or anything and automated spammers should still get banned. 100% agree about your stance on admins.

devchat.go
```diff
@@ -183,15 +188,22 @@ func newUser(s ssh.Session) *user {
 	pty, winChan, _ := s.Pty()
 	w := pty.Window
 	host, _, _ := net.SplitHostPort(s.RemoteAddr().String()) // definitely should not give an err
-	hash := sha256.New()
-	hash.Write([]byte(host))
+	ip_hash := sha256.New()
```
GitOpsBot, line 191: warning: don't use underscores in Go names; var ip_hash should be ipHash (golint)

devchat.go
```diff
@@ -183,15 +188,22 @@ func newUser(s ssh.Session) *user {
...
+	config_hash := sha256.New()
```
GitOpsBot, line 192: warning: don't use underscores in Go names; var config_hash should be configHash (golint)

GitOpsBot (commented):
# GoGitOps Review

## Grade: A+ (95.0%)

### Files: 7

### Issues: 3

### gofmt: 100%

### go_vet: 100%

### golint: 85%

	devchat.go
		Line 191: warning: don't use underscores in Go names; var ip_hash should be ipHash (golint)
	devchat.go
		Line 192: warning: don't use underscores in Go names; var config_hash should be configHash (golint)
### gocyclo: 71%

	util.go
		Line 171: warning: cyclomatic complexity 29 of function devbotChat() is high (> 15) (gocyclo)
	colors.go
		Line 98: warning: cyclomatic complexity 19 of function getStyle() is high (> 15) (gocyclo)
### ineffassign: 100%

### license: 100%

### misspell: 100%




TommyPujol06:
Alright, so shall I remove the IP based ID then?

quackduck:
Yes please! Also use the Go convention for naming variables.

TommyPujol06:
> Yes please! Also use the Go convention for naming variables.

Both done.

devchat.go
```diff
@@ -112,6 +112,10 @@ func main() {
 		}()
 		u.repl()
 	})
+	publicKeyOption := ssh.PublicKeyAuth(func(ctx ssh.Context, key ssh.PublicKey) bool {
+		return true // allow all keys, or use ssh.KeysEqual() to compare against known keys
+	})
+
```
quackduck, line 118:
```suggestion
```

devchat.go
```diff
@@ -131,7 +135,7 @@ func main() {
 			}
 		}
 	}()
-	err = ssh.ListenAndServe(fmt.Sprintf(":%d", port), nil, ssh.HostKeyFile(os.Getenv("HOME")+"/.ssh/id_rsa"))
+	err = ssh.ListenAndServe(fmt.Sprintf(":%d", port), nil, ssh.HostKeyFile(os.Getenv("HOME")+"/.ssh/id_rsa"), publicKeyOption)
```
quackduck, line 138:
```suggestion
	err = ssh.ListenAndServe(fmt.Sprintf(":%d", port), nil, ssh.HostKeyFile(os.Getenv("HOME")+"/.ssh/id_rsa"), func(ctx ssh.Context, key ssh.PublicKey) bool {
		return true // allow all keys, or use ssh.KeysEqual() to compare against known keys
	})
```

devchat.go
```diff
@@ -327,6 +336,7 @@ func (u *user) pickUsername(possibleName string) (ok bool) {
 	var err error
 	for {
 		if possibleName == "" {
+			u.writeln("", "No username chosen. Pick one:")
```
quackduck, line 339:
```suggestion
```

commands.go
```diff
@@ -38,7 +38,7 @@ var (
 		{"rest", commandsRestCMD, "", "Uncommon commands list"}}
 	cmdsRest = []cmd{
 		{"people", peopleCMD, "", "See info about nice people who joined"},
-		{"id", idCMD, "<user>", "Get a unique ID for a user (hashed IP)"},
+		{"id", idCMD, "<user>", "Get a unique ID for a user (hashed IP | SSH Public key)"},
```
quackduck, line 41:
```suggestion
		{"id", idCMD, "<user>", "Get a unique ID for a user (hashed key)"},
```

quackduck:
This is great, thank you so much Tommy

PR #62, opened 2021-11-16
Improved background handling.

Arkaeriit:
It is possible to reset one's background
and to combine a special color and a
background.


quackduck:
Looks good! Thanks Ark

PR #64, opened 2021-11-30
Fixed bug with color random.

Arkaeriit:
When the color had no red, the converted
number was 2 char longs which made the
rest of the getStyle function unable to parse
it. Changing Atoa to Sprintf means that I can
enforce to have leading 0 when the number is
only two digit long, which mean that it will
be parsed correctly.


quackduck:
Thank you :)

PR #66, opened 2022-02-05
Clears empty rooms

Arkaeriit:
When a user left a room where they were alone in, the room was lest open.
This PR fixes that and ensure that there is no empty rooms.

### DIFF
--- a/devchat.go
+++ b/devchat.go
@@ -402,7 +402,7 @@ func (u *user) changeRoom(r *room) {
 	}
 	u.room.users = append(u.room.users, u)
 	u.room.broadcast(devbot, u.name+" has joined "+blue.Paint(u.room.name))
-	cleanupRooms();
+	cleanupRooms()
 }
 
 func (u *user) repl() {
--- a/devchat.go
+++ b/devchat.go
@@ -281,11 +281,9 @@ func newUser(s ssh.Session) *user {
 
 // Deletes any empty room. Checks all rooms so that it works even when the
 // user left.
-func cleanupRooms() {
-	for _, r := range rooms {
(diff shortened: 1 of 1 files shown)


devchat.go
```diff
@@ -279,6 +279,16 @@ func newUser(s ssh.Session) *user {
 	return u
 }
 
+// Deletes any empty room. Checks all rooms so that it works even when the
+// user left.
+func cleanupRooms() {
+	for _, r := range rooms {
```
quackduck, line 285: We only need to check the room this user left, so a loop through all rooms is not necessary. I would have this function take in a room argument to check.

devchat.go
```diff
@@ -289,6 +299,7 @@ func (u *user) close(msg string) {
 			u.room.broadcast(devbot, u.name+" stayed on for "+printPrettyDuration(time.Since(u.joinTime)))
 		}
 	})
+	cleanupRooms()
```
quackduck, line 302: This should be moved into the `closeOnce.Do` function

quackduck:
Thank you! I _was_ wondering what had been causing the empty rooms lol

Arkaeriit:
I fixed the problems you raised.

devchat.go
```diff
@@ -279,6 +279,14 @@ func newUser(s ssh.Session) *user {
 	return u
 }
 
+// Deletes any empty room. Checks all rooms so that it works even when the
+// user left.
```
quackduck, line 283:
```suggestion
// cleanupRoom deletes a room if it's empty and isn't the main room
```

quackduck:
Perfect, thank you so much!

PR #69, opened 2022-02-07
Fixed the warning raised by golint.

Arkaeriit:
This does not changes any behaviors but
it will please goint that conplains at
every pull requests.

### DIFF
--- a/util.go
+++ b/util.go
@@ -36,7 +36,7 @@ func getAdmins() []string {
 		return []string{}
 	}
 	ids := make([]string, 0, len(adminsList))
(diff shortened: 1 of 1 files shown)


quackduck:
While writing that code I actually included that on purpose to emphasize that the notes field is not being used, so maybe disable the linter warning on that line? Or is it clear enough already that the notes field is not being used? What do you think?

Arkaeriit:
I feel like naming the variable `id` is enough to make us understand that you only needed the key of the table. Does that makes sense to you?

PR #72, opened 2022-02-12
Don't panic on file not found.

TommyPujol06:
This commit makes it so that we can access bans in O(1) constant time.
It also allows the program to run if the twitter or slack files are not found.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -367,7 +367,6 @@ func listBansCMD(_ string, u *user) {
 	msg := "Printing bans by ID:  \n"
 	for i := 0; i < len(bans); i++ {
 		msg += cyan.Cyan(strconv.Itoa(i+1)) + ". " + bans[i].ID + "  \n"
-		i++
 	}
 	u.room.broadcast(devbot, msg)
 }
--- a/commands.go
+++ b/commands.go
@@ -365,8 +365,10 @@ func nickCMD(line string, u *user) {
 
 func listBansCMD(_ string, u *user) {
 	msg := "Printing bans by ID:  \n"
-	for i := 0; i < len(bans); i++ {
-		msg += cyan.Cyan(strconv.Itoa(i+1)) + ". " + bans[i].ID + "  \n"
+	i := 0
+	for ban := range bans {
+		msg += cyan.Cyan(strconv.Itoa(i+1)) + ". " + ban.ID + "  \n"
+		i++
 	}
 	u.room.broadcast(devbot, msg)
 }
@@ -376,11 +378,11 @@ func unbanCMD(toUnban string, u *user) {
 		u.room.broadcast(devbot, "Not authorized")
 		return
 	}
-	for i := 0; i < len(bans); i++ {
-		if bans[i].ID == toUnban || bans[i].Addr == toUnban { // allow unbanning by either ID or IP
-			u.room.broadcast(devbot, "Unbanned person: "+bans[i].ID)
-			// remove this ban
-			bans = append(bans[:i], bans[i+1:]...)
+
+	for ban := range bans {
+		if ban.ID == toUnban || ban.Addr == toUnban { // allow unbanning by either ID or IP
+			u.room.broadcast(devbot, "Unbanned person: " + ban.ID)
+			delete(bans, ban)
 		}
 	}
 	saveBans()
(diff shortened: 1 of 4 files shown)


quackduck:
Map access is only better than slice search at high numbers. I'll run a modified version of https://gist.github.com/JelteF/1251f180966eb974d7732ea6c3d3b7ff to see if using a `map[ban]bool‘ really is better than a `[]ban` at our scale (< ~20 bans at any time) Another thing is that `map[ban]bool` implies that bans can be stored with the bool set to false, which we do not want. `[]ban` makes it absolutely clear that we will only store the bans, so it's simpler to understand. I like the Twitter and Slack changes, but I would modify them to return a no-op result and a message indicating that the files are not present instead of carrying on like the API keys are empty (which will cause confusing errors).

CodeLongAndProsper90:
You can already set `DEVZAT_OFFLINE=1` to run offline

quackduck:
Yes, what about it?

CodeLongAndProsper90:
Shouldn't that A) be split into another commit, and B) just say 'set DEVZAT_OFFLINE to run without connecting to external services'?

quackduck:
@TommyPujol06 if the file isn't found, you can set the global variable `offline` to `true` and it should do everything needed (except print a message that the file doesn't exist)

TommyPujol06:
> Map access is only better than slice search at high numbers. I'll run a modified version of https://gist.github.com/JelteF/1251f180966eb974d7732ea6c3d3b7ff to see if using a `map[ban]bool` really is better than a `[]ban` at our scale (< ~20 bans at any time) Another thing is that `map[ban]bool` implies that bans can be stored with the bool set to false, which we do not want. `[]ban` makes it absolutely clear that we will only store the bans, so it's simpler to understand. I like the Twitter and Slack changes, but I would modify them to return a no-op result and print a message indicating that the files are not present instead of carrying on like the API keys are empty (which will cause confusing errors).

Do you think that documenting this behaviour would be enough, or shall I revert the change and keep the O(n) code since we know that for our use case it will be fast enough?

> @TommyPujol06 if the file isn't found, you can set the global variable offline to true and it should do everything needed (except print a message that the file doesn't exist)

Now I made it so that a message is printed and offline mode is set.

quackduck:
> Do you think that documenting this behaviour would be enough, or shall I revert the change and keep the O(n) code since we know that for our use case it will be fast enough?

Keep the O(n) code.

> Now I made it so that a message is printed and offline mode is set.

I just realized that setting offline would disable both slack and twitter if either file is not found, so I would instead replace `offline` with two new variables: `offlineSlack`, `offlineTwitter`. Also, the check for offline happens before you set it, so move the new code before the check. For example, over here: https://github.com/quackduck/devzat/blob/6c82f969ffe9321c0982ce4151246e0a2cf710ff/slack.go#L57

quackduck:
After you revert the map changes, if you'd like, I can help implement the changes to `offline`.

TommyPujol06:
All right, I added a slack and twitter offline as well as a global offline that adds backwards compatibility.

quackduck:
This looks great! I would just replace `errors.Is(err, os.ErrNotExist)` with [`os.IsNotExist()`](https://pkg.go.dev/os#IsNotExist) and invert the ifs like so for readability:

```diff
-	if !errors.Is(err, os.ErrNotExist) {
-		panic(err)
-	} else {
-		offlineSlack = true
-		l.Println("Did not find slackAPI.txt. Enabling offline mode.")
-	}
+	if errors.Is(err, os.ErrNotExist) {
+		offlineSlack = true
+		l.Println("Did not find slackAPI.txt. Enabling offline mode.")
+	} else {
+		panic(err)
+	}
```
(easier to parse "if the err is a `not exist error`, then handle it, otherwise panic" than "if the err is not a `not exist error` then panic, otherwise handle it")

TommyPujol06:
> This looks great! I would just replace `errors.Is(err, os.ErrNotExist)` with [`os.IsNotExist()`](https://pkg.go.dev/os#IsNotExist) and invert the ifs like so for readability:
> 
> ```diff
> -	if !errors.Is(err, os.ErrNotExist) {
> -		panic(err)
> -	} else {
> -		offlineSlack = true
> -		l.Println("Did not find slackAPI.txt. Enabling offline mode.")
> -	}
> +	if errors.Is(err, os.ErrNotExist) {
> +		offlineSlack = true
> +		l.Println("Did not find slackAPI.txt. Enabling offline mode.")
> +	} else {
> +		panic(err)
> +	}
> ```
> 
> (easier to parse "if the err is a `not exist error`, then handle it, otherwise panic" than "if the err is not a `not exist error` then panic, otherwise handle it")

Done.

quackduck:
Thank you so much! Merging ✅

PR #81, opened 2022-03-16
Devbot notify about users changing nicknames.

Arkaeriit:
When a user changed their nick to an already used
nick, there was no info for the other users to
see the final name they choosen.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -416,12 +416,7 @@ func idCMD(line string, u *user) {
 }
 
 func nickCMD(line string, u *user) {
-	oldName := u.name
-	showedPrompt, _ := u.pickUsername(line)
-	if showedPrompt {
-		u.room.broadcast(devbot, oldName+" have been renamed to "+u.name+".")
-	}
-	return
+	u.pickUsernameExplicit(line)
 }
 
 func listBansCMD(_ string, u *user) {
--- a/commands.go
+++ b/commands.go
@@ -416,10 +416,10 @@ func idCMD(line string, u *user) {
 }
 
 func nickCMD(line string, u *user) {
-	old_name := u.name
+	oldName := u.name
 	showedPrompt, _ := u.pickUsername(line)
 	if showedPrompt {
-		u.room.broadcast(devbot, old_name+" have been renamed to "+u.name+".")
+		u.room.broadcast(devbot, oldName+" have been renamed to "+u.name+".")
 	}
 	return
 }
--- a/commands.go
+++ b/commands.go
@@ -416,7 +416,7 @@ func idCMD(line string, u *user) {
 }
 
 func nickCMD(line string, u *user) {
-	u.pickUsernameExplicit(line)
+	u.pickUsername(line)
 }
 
 func listBansCMD(_ string, u *user) {
--- a/commands.go
+++ b/commands.go
@@ -417,8 +417,10 @@ func idCMD(line string, u *user) {
 
 func nickCMD(line string, u *user) {
 	old_name := u.name
-	u.pickUsername(line)
-	u.room.broadcast(devbot, old_name+" have been renamed to "+u.name+".")
+	showedPrompt, _ := u.pickUsername(line)
+	if showedPrompt {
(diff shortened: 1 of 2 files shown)


quackduck:
What if we handle just that special case in `pickUsername`?

Arkaeriit:
I think it would be a bit more noisy as `pickUsername` is already called when a new user log in.

quackduck:
No I mean detecting that the user initially chose a duplicate name and only then printing out this message, otherwise the message shows every time.

Arkaeriit:
I get what you are saying and I will implement it. But I think that we should not change the behavior of `pickUsername` as the way it is used for new users us really smooth.

commands.go
```diff
@@ -411,7 +411,11 @@ func idCMD(line string, u *user) {
 }
 
 func nickCMD(line string, u *user) {
-	u.pickUsername(line)
+	old_name := u.name
```
GitOpsBot, line 414: warning: don't use underscores in Go names; var old_name should be oldName (golint)

GitOpsBot (commented):
# GoGitOps Review

## Grade: A+ (95.0%)

### Files: 7

### Issues: 2

### gofmt: 100%

### go_vet: 100%

### golint: 85%

	commands.go
		Line 414: warning: don't use underscores in Go names; var old_name should be oldName (golint)
### gocyclo: 71%

	util.go
		Line 188: warning: cyclomatic complexity 29 of function devbotChat() is high (> 15) (gocyclo)
	commands.go
		Line 79: warning: cyclomatic complexity 17 of function runCommands() is high (> 15) (gocyclo)
### ineffassign: 100%

### license: 100%

### misspell: 100%




commands.go
```diff
@@ -411,7 +411,11 @@ func idCMD(line string, u *user) {
 }
 
 func nickCMD(line string, u *user) {
-	u.pickUsername(line)
+	oldName := u.name
+	showedPrompt, _ := u.pickUsername(line)
+	if showedPrompt {
+		u.room.broadcast(devbot, oldName+" have been renamed to "+u.name+".")
```
quackduck, line 417: This should be handled right inside `pickUsername()`. It’s that function’s responsibility to change the username and do whatever is needed to properly change it.

Arkaeriit: The `pickUsername` function already handle the change of the name well. But this shouldn't be its responsibility to broadcast info about the rename as the `pickUsername` is also used for new users where we shouldn't try to broadcast anything in the event of a nickname conflict.

GitOpsBot (commented):
All good!

commands.go
quackduck: Good point

quackduck: We can solve the same way we solved the close-silently issue: make a new function that does the name change and all the prompts quietly, and then make `pickUsername()` a wrapper around that function that checks if the changed name is not the same as `possibleName` and if so, prints the notification. The new user function can then call the quiet function, and all other usages of `pickUsername()` will stay the same. If you agree to this, I can implement it if you'd like.

Arkaeriit: If you want to do that you can go for it or ask me to do it if you prefer. But isn't that somewhat similar to what I already made ? Indeed, I have the `pickUsername()` function that does the quiet processing and the `nickCMD()` function does the loud processing.

quackduck: It would not be apparent why other usages call `nickCMD` instead of `pickUsername`, which has a more appropriate name. Also, this could just be me, but calling `cmd` functions in non-cmd contexts feels uncomfortable. 

Another solution is to add a boolean flag specifying whether to be quiet or not, but that seems clunky.

Arkaeriit: I now understand what you mean. I will implement it.

Arkaeriit: It is done.

quackduck: Thanks

GitOpsBot (commented):
All good!

quackduck:
LGTM! Is there anything you want to change?

quackduck:
Okay Arkaeriit said it’s good to merge on Discord. Thanks! I’m merging it.

PR #98, opened 2022-03-23
[feat] Add github action to verify build compiles

polyrain:
Small PR to add a action that makes sure a PR must build. We can enforce this as a requirement for main by changing a repo setting. https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests


Arkaeriit:
Good idea. Could it be possible to add the [GitOpsBot](https://github.com/GitOpsBot) check as well. It might take a bit more work as we want to ignore the gocyclio metric.

GitOpsBot (commented):
All good!

quackduck:
We could add a `golangci-lint` step, which is probably what the bot uses.

polyrain:
If we want to do this, then we should[ add the file they recommend here](https://github.com/marketplace/actions/run-golangci-lint)

quackduck:
Sounds good 👍

polyrain:
Added

quackduck:
Ahh, we might have to add some of those linter-disabling comments in all the places it complains right now. Or maybe we can just disable errcheck and deadcode (they should be found in code review anyway?). 

Also, does this check if the code is formatted correctly? (We could add a workflow that automatically runs `go fmt` and commits the changes)

polyrain:
> Ahh, we might have to add some of those linter-disabling comments in all the places it complains right now. Or maybe we can just disable errcheck and deadcode (they should be found in code review anyway?). 

Up to you, I think the more we disable from the linter the less powerful it becomes but we can also make this not a required workflow (ie it tells the user these issues exist but it doesn't enforce anything)

> 
> Also, does this check if the code is formatted correctly? (We could add a workflow that automatically runs `go fmt` and commits the changes)

It doesn't current check if it's formatted, I think it might be more useful to just run the formatter and let the user know if it's formatted or not since auto committing stuff is always a bit concerning

quackduck:
Alright, we can have it become not-required and I'll add the comments where necessary.

GitOpsBot (commented):
# GoGitOps Review

## Grade: A+ (96.6%)

### Files: 7

### Issues: 2

### gofmt: 100%

### go_vet: 100%

### golint: 100%

### gocyclo: 71%

	util.go
		Line 189: warning: cyclomatic complexity 29 of function devbotChat() is high (> 15) (gocyclo)
	commands.go
		Line 83: warning: cyclomatic complexity 17 of function runCommands() is high (> 15) (gocyclo)
### ineffassign: 100%

### license: 100%

### misspell: 100%




quackduck:
I made it so it only errors out on newly introduced issues, so now PRs should be fine.

PR #101, opened 2022-03-28
wip refactor

gravestench:
**still WIP, need to work more on decoupling the commands from the room / server**

Hi there. I felt like the project was a bit messy, so I've tried to clean it up as best as i could. I know that I've changed a lot of stuff, but it was all done in order to bring some coherence to the project as a whole. One of the main things I've tried to accomplish in this refactor is to adopt the standard project layout, which places build targets inside of `cmd`, and reusable code inside of `pkg`. Unsurprisingly, the `main` function has been moved out of the repo root and into `cmd/devzat-server`, which is the build target for the chat server.

### adopting standard project layout
* see [here](https://github.com/golang-standards/project-layout)
* moved `room` and `user` types into `pkg/models`
* moved colors stuff into `pkg/colors` and declared a `Formatter` type, all color funcs are now methods
* moved server main func out of repo root, into `cmd/devzat_server`

### general changes
* un-cuddled code
* reduced logic where possible
* removed inline string concatenation
* removed global variables
* bubbled up errors where it made sense
* removed `init` functions
* moved global state into object instances
* decoupled the code as much as possible

### changes to commands
* **CURRENTLY WIP** -- moving all commands into their own subdirs inside of `pkg/commands/`
* switching to server-wide command registry implementation
* added `CommandManager` type, stores what used to be global-state
* moved command declarations into their own modules in `pkg/commands/*`


quackduck:
Thank you for making this PR! However, this PR makes too many changes for me to review all of them effectively, and for you to keep updated with new functionality as it is added. This also changes behavior in a lot of places, which should definitely be a separate PR from a refactor one. Please separate this PR into multiple ones with smaller scopes and single logical changes. Thanks!

gravestench:
Nah

PR #108, opened 2022-04-07
Major refactor

gravestench:
hi, im still decoupling everything, but it's starting to look _pretty fuckin nice_.

image

would you look at that? if you were wondering about rooms, or users, or server, or bots, you can just wander around and likely find what you're looking for.

I mean, look:
image

Yeah, the commands are each in their own subdirs. They're pretty close to being totally uncoupled from the rest of the codebase, already! There's still some work to do here, but it's mostly done as of now.

Here is server init, in `cmd/devzat-server/main.go`:
image

I mean, look at that, the code is almost fuckin boring, now!

and here's the underlying server init logic (not the special shit inside of `cmd`):
```golang
func (s *Server) Init() error {
	s.chatHistory.init()
	s.spamManagement.init()
	s.commandRegistry.init()

	if err := s.settings.init(); err != nil {
		return fmt.Errorf("could not init server settings: %s", err)
	}

	if err := s.roomManagement.init(s); err != nil {
		return fmt.Errorf("could not init rooms: %s", err)
	}

	if err := s.serverLogs.init(); err != nil {
		return fmt.Errorf("could not init server logging: %s", err)
	}

	if err := s.banManagement.init(s); err != nil {
		return fmt.Errorf("could not init server bans: %s", err)
	}

	b := &bot.DevBot{}
	b.SetRoom(s.mainRoom)
	if err := b.Init(); err != nil {
		return fmt.Errorf("there was an error initializing the bot: %v", err)
	}

	s.startupTime = time.Now()

	s.loadTwitterClient()

	return nil
}
```

and here, we have the bot response rule declarations, but with a builder pattern:
image

I'm gonna put all the shit inside of a json settings file for the bot.

I've also changed the server/rooms such that:
* rooms can have their own admins
* rooms can have their own bots

And I already know that there is no fuckin way you're going to merge this, but I'm pretty sure that it's going to be impossible for you to port anything over to your codebase because it's just not feasible given the implementation, and you likely don't want to.

Anyways, have a good evening/afternoon/morning/whatever, and i'll smell ya later.


quackduck:
what a legend

sector-f:
This is absolutely beautiful.  

This is what I _wanted_ to do with my fork, but then I decided that it would be too much effort, so I just started from scratch and copypasted as needed instead of trying to maintain feature parity

gravestench:
@sector-f if you want to help me with it, i'll stop amending the single commit and push what i have currently. Then, we can both work on individual tiny things, in individual tiny commits, without getting in each others way.

sector-f:
@gravestench I'm not _inherently_ opposed to that, but also I'm really lazy and already have my own fork that I'm working on, so it would be a mistake for me to say anything that sounds remotely like a promise to help :sweat_smile:

PR #112, opened 2022-04-08
Changing nickname keep the user's color.

Arkaeriit:


quackduck:
This should be handled inside `pickUsername` which is the reason the colors change.

Arkaeriit:
You are right. Maybe pickUsername shouldn't change the color in the first place?

quackduck:
Yes, that's what I've been working on coincidentally.

quackduck:
was fixed in https://github.com/quackduck/devzat/commit/0e0ef1a5c79983ed3ef3dc545f4dac31b79a9098

PR #118, opened 2022-04-19
Update go-away to v1.6.2 (includes Censor bug fix).

awdavies:
See https://github.com/TwiN/go-away/pull/29

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module devzat
 go 1.17
 
 require (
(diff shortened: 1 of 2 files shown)


quackduck:
Thanks for making that bug fix!! This PR makes the same changes as https://github.com/quackduck/devzat/commit/0c113643d11916eb13546af1dcc2472d0c680107 so I’m closing this now.

PR #120, opened 2022-04-19
Add holidays func

twopills:
Hi, i've found the valentines function (very funny), i decided to extends them with holidays check.
This func check find a match with month - day and write a image.
I know this commit it's only for fun, but after all doesn't exist only Valentine's day <3


quackduck:
I love this

PR #121, opened 2022-04-26
Simplify gRPC interface, add regex matching support for listeners

Merlin04:
This is a breaking change; after this is merged, any applications consuming the gRPC API will need to be updated. I've already updated `devzat-node`, and as far as I'm aware no one has written a client in any other language, so we should be good.


quackduck:
Epic, thanks!

PR #123, opened 2022-04-28
Add support for granting RPC tokens within Devzat

Merlin04:


rpc.go
```diff
@@ -312,3 +310,153 @@ func getMiddlewareResult(u *User, line string) string {
...
+}
+
+func generateToken() string {
+	// https://stackoverflow.com/a/59457748
+	b := make([]byte, 32)
+	_, err := rand.Read(b)
+	if err != nil {
+		Log.Fatal("Error generating token: " + err.Error())
+	}
+	token := "dvz@" + hex.EncodeToString(b)
```
quackduck, line 454:
```suggestion
	token := "dvz@" + base64.StdEncoding.EncodeToString(b)
```
use base64 for shorter tokens and clear differentiation from hashes

rpc.go
```diff
@@ -312,3 +310,153 @@ func getMiddlewareResult(u *User, line string) string {
...
+
+func grantTokenCMD(rest string, u *User) {
+	if !auth(u) {
+		u.room.broadcast(Devbot, "Not authorized")
+		return
+	}
+
+	toUser, ok := findUserByName(u.room, rest)
+	if !ok {
+		// fallback to sending the token to the admin running the cmd
```
quackduck, line 435: I would err out at this point and let the admin choose whether to run again with a corrected user arg or without one.

rpc.go
```diff
@@ -312,3 +310,153 @@ func getMiddlewareResult(u *User, line string) string {
...
+	_, err = f.Write(data)
+	if err != nil {
+		error := "Error writing tokens file: " + err.Error()
+		MainRoom.broadcast(Devbot, error)
+		Log.Fatal(error)
+	}
+}
+
+func checkToken(token string) bool {
+	if token == Integrations.RPC.Key {
```
quackduck, line 379: We can remove `Integrations.RPC.Key` completely now (tokens are better)

rpc.go
```diff
@@ -312,3 +310,153 @@ func getMiddlewareResult(u *User, line string) string {
...
+		error := "Error decoding tokens file: " + err.Error()
+		MainRoom.broadcast(Devbot, error)
+		Log.Fatal(error)
+	}
+}
+
+func saveTokens() {
+	f, err := os.Create(Config.DataDir + string(os.PathSeparator) + "tokens.json")
+	if err != nil {
+		Log.Fatal("Error saving tokens file: " + err.Error())
```
quackduck, line 361:
```suggestion
		Log.Println("Error saving tokens file:", err)
```
Don't use `Log.Fatal` anywhere since it brings down the whole server.

quackduck (changes requested):
We will also need to update docs

rpc.go
Merlin04: Do you think it's a good idea to also have something that can be configured by the server owner that no one can revoke or view? I initially removed it but I added it back in thinking that could be useful (also to set the auth token on a plugin before you have your devzat instance deployed)

rpc.go
Merlin04: I designed it to have this behavior so that you can create separate tokens that aren't associated with users (like if you wanted to `grant-token Hack Club Emoji Middleware`), it does have the cost of the possibility of typos not sending the message to the user but you can always just dm it or revoke/reissue the token

rpc.go
quackduck: Oh yes, good point, that could be useful. We can keep it for now.

Merlin04:
> We will also need to update docs

And the docs have been updated!

rpc.go
quackduck: You can detect if the admin wants to specify a user by requiring them to use `@user` like some commands do. Then you can respond to typos and also allow for the `grant-token Hack Club Emoji Middleware` use case.

rpc.go
```diff
@@ -312,3 +310,153 @@ func getMiddlewareResult(u *User, line string) string {
...
+	}
+	for i, t := range Tokens {
+		if shasum(t.Token) == rest {
+			Tokens = append(Tokens[:i], Tokens[i+1:]...)
+			saveTokens()
+			u.writeln(Devbot, "Token revoked!")
+			return
+		}
+	}
+	u.writeln(Devbot, "Token not found.")
```
quackduck, line 424: It's correct to not broadcast hashed tokens but status updates like these are okay to broadcast.

commands.go
```diff
@@ -68,13 +68,25 @@ var (
...
 const (
 	MaxRoomNameLen = 30
 )
 
-func init() {
+func initCMDs() {
+	if Integrations.RPC != nil {
+		MainCMDs = append(MainCMDs, RpcCMDs...)
+		RestCMDs = append(RestCMDs, RpcCMDsRest...)
+	}
```
quackduck, line 89: Move these new slice defs into rpc.go and define a new function (`rpcAddCMDS()`) that is used in commands.go's `init()` to add those slices if RPC is enabled.

quackduck: Resolved

Merlin04:
Looks good to me, feel free to merge!

PR #127, opened 2022-05-18
Added Rustrict into the server

Arkaeriit:
As this PR includes the Rust code into the repository you might not want to use it before I make more testing about making a binary-only lib.
You can still play with it as it is fully functional.


quackduck:
Thanks for making this! As we discussed earlier, having a separate library do the rust integration would be ideal, so this can be a draft PR for now.

Arkaeriit:
I made some research on how to package the Go-ed Rustrict as a library and no solution are goods.
The support for binary library in the Go toolchain is quite poor, we cannot automatically install the library like with source-based Go library. This meas that we would need to either rely on custom and complicated install and build scripts to manage the binary library or we let the user/developer figure it out which is probably the worst solution.
Furthermore, as a binary library needs to be OS and architecture specific, we will probably need to make the build process more complex.
An other alternative would be to bundle .a files in the repository and have a build script choose the right one to use. This is easier than managing the installation of libraries in the user's system but it is really not elegant and, as we distribute not all the source code, a bit out of the scope of the MIT license.
A third option would be to add the Rust source code in the repository as it is in this pull request. This have some issues (not suitable for cross-compilation yet, needs cargo installed...) but I feel as this is not as bad as the two previous ones. Any complicated build/install script can be replaced with a simple makefile.
The last option would be to ditch Restrict and keep go-away which does not works well.

I know my text is a bit confused so if you need clarification, don't be afraid to ask.

quackduck:
Did the steps under "Packaging up" here: https://words.filippo.io/rustgo/ not work?

Arkaeriit:
The solution in the link you provided is the first possibility I have described. The library need to be installed in a non-standard way, which is problematic.

quackduck:
> Well, this was fun.
 But to be clear, rustgo is not a real thing that you should use in production. For example, I suspect I should be saving g before the jump, the stack size is completely arbitrary, and shrinking the trampoline frame like that will probably confuse the hell out of debuggers. Also, a panic in Rust might get weird.

This is what the rustgo author says, so maybe we should consider porting rustrict to go normally instead of trying to link it in.

Arkaeriit:
The packaging proposed by this article is what I described in the first solution. We would need to distribute tarballs of .a file. People would then need to put those .a files in their GOPATH to compile Devzat. This is really not a great way to do that.

Arkaeriit:
> This is what the rustgo author says, so maybe we should consider porting rustrict to go normally instead of trying to link it in.
> 
> It's probably more safe since we're using cgo but it still feels vaguely hacky to me.

As far as I know, what I did was the more standard way to call Rust from Go. I used the FFI of both languages as they were intended to be used.

Arkaeriit:
I think that the state of this PR is the best way we can have Rustrict in Devzat. If you might want to merge it, you should unset it as draft and I could try to work on how to make the CI compatible with it. Alternatively, if you think the profanity filter is not worth having Rust in the codebase, you can close it.

quackduck:
I'm not sure if this is worth integrating into the main branch, but it's definitely something we should let the user have the option of using.

My current plan is to make a new `rustrict` branch in the repo to which you can merge this PR. 

To prevent future merge conflicts (when I try to update the `rustrict` branch with commits from `main`), we could add a new `censor.go` file which would be the only go file that differs in the `main` and `rustrict` branches.

Makefile
```diff
@@ -0,0 +1,29 @@
+# This Makefile is needed to compile the Rust glue layer between Devzat and the
+# Rust library Rustrict. To compile Devzat, simply run `make`. To install it as
+# a Go binary, run `make install`.
+
+all: devzat
+
+GO_SRC := colors.go commands.go config.go games.go main.go rpc.go slack.go twitter.go util.go
```
quackduck, line 7: This would include `censor.go` too, right? Is there a better way to do this: does `*.go` work?

quackduck:
As discussed, after review, the next step is to merge this into the new `rustrict` branch I just made.

Makefile
Arkaeriit: Well spotted. I will fix it.

quackduck:
Thank you so much @Arkaeriit!! This is going to be great.

PR #128, opened 2022-05-29
Moved all censor-related functions in a separate file.

Arkaeriit:
As we talked about, I move what you made with go-away in a separate file. Once you merge this into main, I will do the same in the rustrict branch.


quackduck:
Looks good!

Issue #77, opened 2022-03-16
Server doesn't accept RSA SHA-2 keys.


thepixelmonk:
I'm getting a permission denied error even though I have a public key generated. Any ideas?

```
λ ssh pixelmonk@devzat.hackclub.com
pixelmonk@devzat.hackclub.com: Permission denied (publickey).

λ ls ~/.ssh
id_rsa  id_rsa.pub  known_hosts  known_hosts.old
```

Arkaeriit:
This is mentioned in the Readme, you should use the command `ssh-keygen` to get a public key and be able to log in.

Alternatively, you can log to the port 443 which doesn't require a public key.

thepixelmonk:
> This is mentioned in the Readme, you should use the command `ssh-keygen` to get a public key and be able to log in.

Like I said, I already have a public key, which you can see above. Am I missing something?

thepixelmonk:
Found a hacker news post which I think is relevant. It just doesn't like RSA. https://news.ycombinator.com/item?id=30690988

quackduck:
Yeah this is a known issue with some crypto libs: see https://github.com/golang/crypto/pull/197 , https://github.com/gliderlabs/ssh/issues/145 & https://github.com/charmbracelet/soft-serve/issues/48

quackduck:
This problem hasn't been solved, reopening.

kuquay:
Not sure if this is a related error, but I've also been unable to log on to the server, although only recently, as of the beginning of this week. My error is `Unable to negotiate with 150.136.142.44 port 22: no matching host key type found. Their offer: ssh-rsa`
Generating new keys of both RSA and ED25519 variety does not work, as the server does not offer a (working) host key.

quackduck:
Did you get the same error as @thepixelmonk before this week? I wonder if simply changing the server’s ssh key would fix this.

kuquay:
@quackduck I got no errors before this week.

quackduck:
changing the server key will have all old user clients complaining about a possible man in the middle attack lol

kuquay:
@quackduck If no one can reproduce the issue, it's probably not worth it, I don't want to cause all those issues just because I maybe have something on my end.

quackduck:
No this is a legitimate issue a lot of people face: I have to tell them to use port 443. 

We're mostly waiting on the libs here to add support: https://github.com/quackduck/devzat/issues/77#issuecomment-1072923781

[...]

PR #130, opened 2022-06-23
Private chat

Arkaeriit:
One user wanted to have a private chat with Devzat. I made so that this can optionally be made.

### DIFF
--- a/Admin's Manual.md
+++ b/Admin's Manual.md
@@ -66,6 +66,22 @@ kick <user>
 
 If running these commands makes Devbot complain about authorization, you need to add your ID under the `admins` key in your config file (`devzat-config.yml` by default).
 
+### Enabling a user allowlist
+
+Devzat can use be used as a private chatroom. Add this to your config:
+
+```yaml
+private: true # enable allowlist checking
+allowlist: 
+  272b326d7d5e9a6b1d98a10b453bdc8cc950fc15cae2c2e858e30645c72ae7c0: 'John Doe'
+  ...
+```
+
+The `whitelist` has the same format as the `admins` list. Add the IDs of the allowed user and some info about that user (this is to make IDs easier to identify when editing the config file, and isn't used by Devzat)
+
+All admins are allowed even if their ID is not in the whitelist. So, if everyone on the private server is an admin, a whitelist isn't necessary, just enable private mode.
+
+Message backlog on `#main` is disabled in private chats. Only those logged in at the same time as you can read your messages.
 
 ### Enabling integrations
 
@@ -130,21 +146,4 @@ There are 4 environment variables you can set to quickly disable integrations on
 * `DEVZAT_OFFLINE_SLACK=true` will disable Slack
 * `DEVZAT_OFFLINE_RPC=true` will disable the gRPC server
 * `DEVZAT_OFFLINE=true` will disable all integrations.
-
-### Enabling a user allowlist
-
-Devzat can use be used as a private chatroom. Add this to your config:
-
-```yaml
-private: true # enable allowlist checking
-allowlist: 
-  272b326d7d5e9a6b1d98a10b453bdc8cc950fc15cae2c2e858e30645c72ae7c0: 'John Doe'
-  ...
-```
-
-The `whitelist` has the same format as the `admins` list. Add the IDs of the allowed user and some info about that user (this is to make IDs easier to identify when editing the config file, and isn't used by Devzat)
-
-All admins are allowed even if their ID is not in the whitelist. So, if everyone on the private server is an admin, a whitelist isn't necessary, just enable private mode.
-
-Message backlog on `#main` is disabled in private chats. Only those logged in at the same time as you can read your messages.
 
--- a/Admin's Manual.md
+++ b/Admin's Manual.md
@@ -131,21 +131,20 @@ There are 4 environment variables you can set to quickly disable integrations on
 * `DEVZAT_OFFLINE_RPC=true` will disable the gRPC server
 * `DEVZAT_OFFLINE=true` will disable all integrations.
 
-### Making a private chat
+### Enabling a user allowlist
 
-Devzat have been made to be a public chat-room. But if you want, you can use it as a private, invite-only chat-room. To do so, you should add the following lines in the configuration file:
+Devzat can use be used as a private chatroom. Add this to your config:
 
 ```yaml
-private: true # Enables the Whitelist checking
-whitelist: 
-  272b326d7d5e9a6b1d98a10b453bdc8cc950fc15cae2c2e858e30645c72ae7c0: 'John Doe' # One entry for each person allowed to log in
+private: true # enable allowlist checking
+allowlist: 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -170,13 +170,15 @@ func main() {
...
 	fmt.Printf("Starting chat server on port %d and profiling on port %d\n", Config.Port, Config.ProfilePort)
 	go getMsgsFromSlack()
-	go func() {
-		fmt.Println("Also starting chat server on port", Config.AltPort)
-		err := ssh.ListenAndServe(fmt.Sprintf(":%d", Config.AltPort), nil, ssh.HostKeyFile(Config.KeyFile))
-		if err != nil {
-			fmt.Println(err)
-		}
-	}()
+	if !Config.Private {
```
quackduck, line 173: why do you disable the alternate port?

Admin's Manual.md
```diff
@@ -130,3 +130,19 @@ There are 4 environment variables you can set to quickly disable integrations on
 * `DEVZAT_OFFLINE_SLACK=true` will disable Slack
 * `DEVZAT_OFFLINE_RPC=true` will disable the gRPC server
 * `DEVZAT_OFFLINE=true` will disable all integrations.
+
+### Making a private chat
+
+Devzat have been made to be a public chat-room. But if you want, you can use it as a private, invite-only chat-room. To do so, you should add the following lines in the configuration file:
+
+```yaml
+private: true # Enables the Whitelist checking
```
quackduck, line 139: I would just use the length of the whitelist to check if it's a private chat (len 0 == public, else private)

Admin's Manual.md
```diff
@@ -130,3 +130,19 @@ There are 4 environment variables you can set to quickly disable integrations on
...
+whitelist: 
+  272b326d7d5e9a6b1d98a10b453bdc8cc950fc15cae2c2e858e30645c72ae7c0: 'John Doe' # One entry for each person allowed to log in
+  ...
+```
+
+The `whitelist` entry is a map similar to the `admins` map. The keys are the ID of the allowed user and the value are a string whose content is not used.
+
+Even if someone is in the `admins` list, they would be rejected from a private server if their ID is not in the `whitelist` list.
```
quackduck, line 147: I don't see any downside for allowing all admins regardless of whether they're in the whitelist. Do you see any? If we allow it then people don't have to put admins in two lists.

Arkaeriit: That makes sense. I will fix that.

main.go
Arkaeriit: The alternate port does not checks for ssh keys. I think that this is a good behavior but this should not be kept for private chats.

Admin's Manual.md
Arkaeriit: This might not be a good idea. Something as restrictive as making the server private being behind an implicit settings feels dangerous.
Furthermore, as I followed your advice about allowing admins, a private server might have an empty whitelist.

main.go
quackduck: Okay let's make it check for keys since #77 is now fixed.

Admin's Manual.md
quackduck: hmm. good points. I'm wondering if having it be implicit + having people add one admin id to the whitelist in the case that everyone allowed is an admin is a good system.

main.go
```diff
@@ -325,8 +325,9 @@ func newUser(s ssh.Session) *User {
 	}
 
 	if Config.Private {
-		_, ok := Config.WhiteList[u.id]
-		if !ok {
+		_, okWhitelist := Config.WhiteList[u.id]
+		_, okAdmin := Config.Admins[u.id]
+		if !okAdmin && !okWhitelist {
```
quackduck, line 330:
```suggestion
		if !(okAdmin || okWhitelist) {
```

Admin's Manual.md
```diff
@@ -144,5 +144,6 @@ whitelist:
 
 The `whitelist` entry is a map similar to the `admins` map. The keys are the ID of the allowed user and the value are a string whose content is not used.
 
-Even if someone is in the `admins` list, they would be rejected from a private server if their ID is not in the `whitelist` list.
+If someone is in the `admins` list, it is assumed that they are allowed. They, they would not be rejected from a private server even if their ID is not in the `whitelist` list. 
+If everyone on the private server is an admin, you don't need to have a `whitelist` list as all the ID are already in the `admin` list.
```
quackduck, line 148:
```suggestion
All admins are allowed even if their ID is not in the whitelist. 
Therefore, if everyone on the private server is an admin, a whitelist isn't necessary.
```

main.go
```diff
@@ -354,15 +368,17 @@ func newUser(s ssh.Session) *User {
...
 	}
 
-	if len(Backlog) > 0 {
-		lastStamp := Backlog[0].timestamp
-		u.rWriteln(fmtTime(u, lastStamp))
-		for i := range Backlog {
-			if Backlog[i].timestamp.Sub(lastStamp) > time.Minute {
-				lastStamp = Backlog[i].timestamp
-				u.rWriteln(fmtTime(u, lastStamp))
+	if !Config.Private {
```
quackduck, line 371: we've already refused non-allowed people, right? we can show the allowed people the backlog

config.go
```diff
@@ -16,6 +16,8 @@ type ConfigType struct {
 	KeyFile     string            `yaml:"key_file"`
 	Admins      map[string]string `yaml:"admins"`
 	Censor      bool              `yaml:"censor,omitempty"`
+	Private     bool              `yaml:"private,omitempty"`
+	WhiteList   map[string]string `yaml:"whitelist"`
```
quackduck, line 20: This can be `omitempty`

main.go
Arkaeriit: Are you sure that this would be a good idea, if for some reason some people can use key files, offering an alternative login method would be very handy.

Admin's Manual.md
Arkaeriit: I will do that if you really wants to, but I don't like having such a bug configuration switch be implicit.

main.go
Arkaeriit: If people need private chats, they might need it to discuss sensible manner. In that case, they probably wouldn't want some people reading their writings in the future without them being able to prevent that.

Admin's Manual.md
Arkaeriit: Nice and to the point, thanks.

main.go
Arkaeriit: No objections to that.

config.go
Arkaeriit: Indeed, it is better that way. Nice catch.

main.go
quackduck: good point

Admin's Manual.md
quackduck: you're right, let's keep it explicit

main.go
quackduck: okay let's add a comment that says this ^

quackduck:
thanksssss

PR #133, opened 2022-09-01
Greedy mention

Arkaeriit:
The `@<username>` mentions are greedy so that if `user1` and `user11` are both connected, `@user11` always mentions `user11` and never `user1`.

This PR also introduces unit test that I wrote for the greedy mention feature but that could be used in other contexts.

As of writing this message, the PR is not ready to be merged yet. I am opening it now if you want to review the unit tests @quackduck.

### DIFF
--- a/colors.go
+++ b/colors.go
@@ -134,7 +134,9 @@ func (u *user) changeColor(colorName string) error {
 	//styleBG, _ := getStyle(u.colorBG)
 	//u.name = styleFG.apply(u.name) // fg clears the bg color
 	//u.name = styleBG.apply(u.name) // then re-add bg color if any
-	u.term.SetPrompt(u.name + ": ")
+    if u.term != nil { // TODO: fix this in tester.go
+        u.term.SetPrompt(u.name + ": ")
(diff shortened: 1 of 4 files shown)


Arkaeriit:
I think the PR is now ready to be merged.

quackduck:
I'm unsure whether we need unit tests (testing is nice when functions are mostly pure). Greedy mentions sound great!

Arkaeriit:
Even if we know we won't have 100 % test coverage in Devzat, some unit tests can't hurt. The unit test for the coloring of mentions might help us catch bugs in the future caused by changes in the room or user struct. Furthermore, a framework to test Devzat without spinning up an instance can help a lot when we want to add in new features.

Arkaeriit:
I just added a bit of code to the mention function and the unit test saved me a lot of time. I am sure that some similar situation will come in the future.

quackduck:
TBH the LOC the tests were adding was why I was a little hesitant. I think we can try out tests and they could be useful. Thanks!

PR #135, opened 2022-09-23
Add https:// before address of the line remover.

Arkaeriit:
This will help terminal notice that this is a
link that could be clicked.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -595,7 +595,7 @@ Interesting features:
 * Built in Tic Tac Toe and Hangman! Run tic or hang <word> to start new games.
 * Emoji replacements! \:rocket\: => :rocket: (like on Slack and Discord)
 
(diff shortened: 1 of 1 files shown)


commands.go
```diff
@@ -595,7 +595,7 @@ Interesting features:
 * Built in Tic Tac Toe and Hangman! Run tic or hang <word> to start new games.
 * Emoji replacements! \:rocket\: => :rocket: (like on Slack and Discord)
 
-For replacing newlines, I often use bulkseotools.com/add-remove-line-breaks.php.
+For replacing newlines, I often use https://bulkseotools.com/add-remove-line-breaks.php.
```
quackduck, line 598:
```suggestion
For replacing newlines, I often use https\://bulkseotools.com/add-remove-line-breaks.php.
```

PR #146, opened 2023-01-30
Improve admins command

Arkaeriit:
As we have some inactive users with white-space names, having commands that can use IDs instead of names is quite valuable.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -542,9 +542,8 @@ func banUser(banner string, victim *User, dur time.Duration) {
 		}(victim.id) // evaluate id now, call unban with that value later
 	}
 	id := victim.id
-	r := victim.room
 	victim.ban(victim.Name + " has been banned by " + banner)
-	kickId(id, banner, r)
+	kickId(id, banner)
 }
 
 func kickCMD(line string, u *User) {
@@ -557,15 +556,17 @@ func kickCMD(line string, u *User) {
 		u.room.broadcast(Devbot, "Not authorized")
 		return
 	}
-	kickId(victim.id, u.Name, u.room)
+	kickId(victim.id, u.Name)
 }
 
 // Kicks all users using the given ID in the given room
-func kickId(id string, kicker string, r *Room) {
-	victim, ok := findUserById(r, id)
-	for ok {
-		victim.close(victim.Name + Red.Paint(" has been kicked by ") + kicker)
-		victim, ok = findUserById(r, id)
+func kickId(id string, kicker string) {
+	for _, r := range Rooms {
+		victim, ok := findUserById(r, id)
+		for ok {
+			victim.close(victim.Name + Red.Paint(" has been kicked by ") + kicker)
+			victim, ok = findUserById(r, id)
+		}
 	}
 }
 
--- a/commands.go
+++ b/commands.go
@@ -537,9 +537,7 @@ func banCMD(line string, u *User) {
 			unbanIDorIP(id)
 		}(victim.id) // evaluate id now, call unban with that value later
 	}
-	id := victim.id
 	victim.ban(victim.Name + " has been banned by " + u.Name)
-	kickId(id, u.Name)
 }
 
 func kickCMD(line string, u *User) {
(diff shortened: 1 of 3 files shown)


commands.go
```diff
@@ -514,7 +514,7 @@ func banCMD(line string, u *User) {
 		u.room.broadcast(Devbot, "Which user do you want to ban?")
 		return
 	}
-	victim, ok := findUserByName(u.room, split[0])
+	victim, ok := findUserByIdOrName(u.room, split[0])
```
quackduck, line 517: what is the use of this?

Arkaeriit: I feel like it is better to let the admin have the choice. If someone manages in the future to have a malformed name, we would be happy to kick using IDs.

Arkaeriit:
I feel like the PR is ready to be merged, there is all the features I wanted to add.

commands.go
quackduck: That means that this feature is only useful when the name cleaning function has a bug. I think this is a very small use case and that the bug's root cause should be fixed instead when it arises.

Arkaeriit: I totally agree with you on that one, we should always strive to fix the root of an issue. Yet, if such an issue does arise, having an admin do some cleaning on the chatroom before you have time to fix the bug could come in handy.

If you really don't like that feature, I can take it out. The rest of the PR does make sense even without it.

quackduck: I think we can take it out. Super small use case ~~+ I think I fixed all the name bugs already~~

Arkaeriit: Done!

commands.go
```diff
@@ -553,7 +558,16 @@ func kickCMD(line string, u *User) {
 		u.room.broadcast(Devbot, "Not authorized")
 		return
 	}
-	victim.close(victim.Name + Red.Paint(" has been kicked by ") + u.Name)
+	kickId(victim.id, u.Name, u.room)
+}
+
+// Kicks all users using the given ID in the given room
+func kickId(id string, kicker string, r *Room) {
```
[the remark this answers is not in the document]
Arkaeriit: This is a very good suggestion. I will implement it.

Kicking users in the username picking stage would take a bit more work but I can implement it if you feel like we need it.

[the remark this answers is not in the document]
quackduck: i think the ban evade can be fixed by simply changing when the ban check occurs or repeating it. not a costly op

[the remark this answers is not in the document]
Arkaeriit: The kicking on all rooms is done.

Do you know how to reproduce Epic's bug? I can't find how to fix it if I can't do it.

commands.go
```diff
@@ -521,20 +521,29 @@ func banCMD(line string, u *User) {
...
 		if err != nil {
 			u.room.broadcast(Devbot, "I couldn't parse that as a duration")
 			return
 		}
-		victim.ban(u.Name + " has been banned by " + u.Name + " for " + dur.String())
+	}
+	banUser(u.Name, victim, dur)
+}
+
+func banUser(banner string, victim *User, dur time.Duration) {
```
quackduck, line 536: I think we can merge `banUser`'s functionality directly into `user.ban`

commands.go
```diff
@@ -521,20 +521,25 @@ func banCMD(line string, u *User) {
...
+		victim.ban(victim.Name + " has been banned by " + u.Name + " for " + dur.String())
 		go func(id string) {
 			time.Sleep(dur)
 			unbanIDorIP(id)
 		}(victim.id) // evaluate id now, call unban with that value later
-		return
 	}
-	victim.ban(u.Name + " has been banned by " + u.Name)
+	id := victim.id
+	victim.ban(victim.Name + " has been banned by " + u.Name)
```
quackduck, line 541: Maybe the `user.ban` func should always use `kickID`

Arkaeriit: What do you mean?

quackduck: since the ban function kicks the banned user, it should maybe use the new kick command to kick all users with the same id as the banned user

Arkaeriit: As of [7bd2c1a](https://github.com/quackduck/devzat/pull/146/commits/7bd2c1aeeca1d735403a3fd2e09de8ad75d966ef), it is what is done at line 542 of commands.go.

quackduck:
Thanks for the PR!!!

PR #147, opened 2023-01-30
Sanitize name on first login.

Arkaeriit:
Before that user could use a new key to bypass
the name sanitation.


quackduck:
this bug has already been fixed :|

PR #149, opened 2023-02-13
I was asked by the server to slap the devlopper

Arkaeriit:
Instead, I will just do that PR.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -542,9 +542,8 @@ func banUser(banner string, victim *User, dur time.Duration) {
 		}(victim.id) // evaluate id now, call unban with that value later
 	}
 	id := victim.id
-	r := victim.room
 	victim.ban(victim.Name + " has been banned by " + banner)
-	kickId(id, banner, r)
+	kickId(id, banner)
 }
 
 func kickCMD(line string, u *User) {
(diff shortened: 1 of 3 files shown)


quackduck:
There was a simpler solution: simply using the longer form of a slice range: 
```
for j := 0; j < len(Rooms[i].users); j++ {
```

Arkaeriit:
> There was a simpler solution: simply using the longer form of a slice range:
> 
> ```go
> for j := 0; j < len(Rooms[i].users); j++ {
> ```

This might not crash but it is not correct either. If your user list is like:
```
goodUser | userToBan | otherUserToBan | otherGoodUser
```
When `j = 1`, `userToBan` will be kicked, then the list will be like:
```
goodUser | otherUserToBan | otherGoodUser
```
At the next iteration, j will be equal to 2 and point to `otherGoodUser`. Thus, `otherUserToBan` will have escaped being kicked.

I think the best course of action would be to set up some unit tests for the ban system.

Arkaeriit:
> > There was a simpler solution: simply using the longer form of a slice range:
> > ```go
> > for j := 0; j < len(Rooms[i].users); j++ {
> > ```
> 
> This might not crash but it is not correct either. If your user list is like:
> 
> ```
> goodUser | userToBan | otherUserToBan | otherGoodUser
> ```
> 
> When `j = 1`, `userToBan` will be kicked, then the list will be like:
> 
> ```
> goodUser | otherUserToBan | otherGoodUser
> ```
> 
> At the next iteration, j will be equal to 2 and point to `otherGoodUser`. Thus, `otherUserToBan` will have escaped being kicked.
> 
> I think the best course of action would be to set up some unit tests for the ban system.

After some tests I wan wrong, the decrementing of j at the end of the loop does ensure that the code works properly. Good job.

PR #150, opened 2023-02-14
Bans unit tests

Arkaeriit:
This PR adds unit test for the banning behavior.

### DIFF
--- a/devzat_test.go
+++ b/devzat_test.go
@@ -4,10 +4,14 @@
 package main
 
 import (
+	"io"
+	"net"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/acarl005/stripansi"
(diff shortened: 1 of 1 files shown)


Arkaeriit:
Good news, the test works.

Arkaeriit:
The fact that this PR adds a mock for the session interface means that we can use the code I added to test a lot of other behaviors.

quackduck:
What other behaviors do you think would be useful to test in this way?

Arkaeriit:
> What other behaviors do you think would be useful to test in this way?

I wanted for some times to change the config yaml to let admins have more than one keys. It would be very convenient for Elitedino or Epic. But this kind of heavy change would benefit for tests to do TDD.

PR #154, opened 2023-03-20
Alternate Display for Username in Discord

PPTide:
Edit: This requires the bot to have the permissions manage Webhooks (I think I just gave it Admin xD)


PPTide:
Screenshot 2023-03-20 at 21 34 56

It looks like this.

Idk if this fits the theme of devzat but I was bored :)

quackduck:
That's such a cool idea! I think this would be great as an optional thing in the discord integration config. It takes up more vertical space but I see the appeal.

PPTide:
I want to add a pfp depending on the color of the username and I have 3 idea how to do it but want your input @quackduck ...
- open up a port on devzat to host the images
- add a micro service on another machine/the same machine to generate/host the images
- upload them to another channel on discord and then get a url from there (you can upload images to discord but can't use the in the pfp of a webhook...)

quackduck:
I don't really want to make Devzat, which is an SSH server, _also_ an HTTP server but since this is an optional feature, _maybe_ it's okay. I'm thinking that an implementation could work like this:

1. User sends message
2. Discord integration sets avatar_url to "<server url>/pfps/<description of color>.png"
3. An HTTP server listens for connections to <server url>/pfps and uses the color description to send back a very minimal generated image to Discord

The hard part is keeping the terminal colors and discord colors the same, because the way messages work in Devzat means that they don't encode color information. So, I think we'd have to just make the color the result of some sender-name-based function. (eg. hash sender name, then convert hash into RGB values with math).

Ultimately the end result would be pfps with colors that don't match those in the terminal and are only sender-name-dependent (meaning that if someone changes their name, the color will change too), which is kind of not that big an improvement and now involves a whole HTTP server and PNG generating - all of which presents opportunities for holes and DoS attacks ~~from @TAG-Epic~~. So yeah idk if we should do this. What do you and @Arkaeriit think?

quackduck:
oops, closed by mistake

PPTide:
Added Profile picture colors :)

quackduck:
That's awesome! Could you attach an example demo screenshot?

PPTide:
Profile pictures on Discord compared to Devzat usernames: 
Screenshot 2023-03-24 at 18 32 32

PPTide:
https://api.slack.com/methods/chat.postMessage

This is what I found looking through the slack api documentation maybe there'll be the same problem... i don't really know slack tho...

Arkaeriit:
Can the inverse transform be applied to messages send on Discord and read from SSH? It would be nice if the name of Discord users on SSH mirrored the color of their profile pictures.
But this is probably out of the scope of this PR as this probably require quite a bit of work as far as image processing goes.

PPTide:
> Can the inverse transform be applied to messages send on Discord and read from SSH? It would be nice if the name of Discord users on SSH mirrored the color of their profile pictures.
> 
> But this is probably out of the scope of this PR as this probably require quite a bit of work as far as image processing goes.

Maybe create a new issue?

quackduck:
This is such a cool feature that I think we should make this the default discord mode. Is there any extra set-up needed on the Discord side?

PPTide:
> This is such a cool feature that I think we should make this the default discord mode. Is there any extra set-up needed on the Discord side?

It needs "Manage Webhooks" and "Message Content Intent". 
Doesn't need "Send Message" anymore.

quackduck:
> > This is such a cool feature that I think we should make this the default discord mode. Is there any extra set-up needed on the Discord side?
> 
> 
> 
> It needs "Manage Webhooks" and "Message Content Intent". 
> 
> Doesn't need "Send Message" anymore. 

Might need an update to the docs in that case

quackduck:
Just tested, seems not to have broken anything!
image

quackduck:
Epic PR, thanks for this

PR #157, opened 2023-04-21
Discard message to bridge if the channels are full

Arkaeriit:
This will prevent the main chat to be stuck
because of a limitation of an external API.


quackduck:
Makes sense, let's just make those fmt.Println's Log.Println()s

PR #161, opened 2023-05-06
Add :wq as an alias of exit

PPTide:
Because my fingers are faster than my Brain :)


quackduck:
Do we really need two vim exit aliases

PPTide:
yes! yes we do... xD

Arkaeriit:
> Do we really need two vim exit aliases

Yes

PPTide:
Thanks for agreeing with me xD

Sent with Spark

PR #162, opened 2023-05-11
User-customizable prompts

Arkaeriit:
If you want to play with it a bit before giving further info.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -42,7 +42,7 @@ var (
 		{"cd", cdCMD, "#room|user", "Join #room, DM user or run cd to see a list"}, // won't actually run, here just to show in docs
 		{"tz", tzCMD, "<zone> [24h]", "Set your IANA timezone (like tz Asia/Dubai) and optionally set 24h"},
 		{"nick", nickCMD, "<name>", "Change your username"},
-		{"prompt", promptCMD, "<prompt>", "Change your promt. You can use \\w, \\W, \\h, and \\u PS1 tags"},
+		{"prompt", promptCMD, "<prompt>", "Change your promt. You can use \\w, \\W, \\t, \\h, and \\u PS1 tags"},
 		{"pronouns", pronounsCMD, "@user|pronouns", "Set your pronouns or get another user's"},
 		{"theme", themeCMD, "<theme>|list", "Change the syntax highlighting theme"},
 		{"rest", commandsRestCMD, "", "Uncommon commands list"}}
@@ -402,6 +402,7 @@ func cdCMD(rest string, u *User) {
 }
 
 func tzCMD(tzArg string, u *User) {
+	defer u.showPrompt()
 	if tzArg == "" {
 		u.Timezone.Location = nil
 		u.room.broadcast(Devbot, "Enabled relative times!")
--- a/commands.go
+++ b/commands.go
@@ -42,7 +42,7 @@ var (
 		{"cd", cdCMD, "#room|user", "Join #room, DM user or run cd to see a list"}, // won't actually run, here just to show in docs
 		{"tz", tzCMD, "<zone> [24h]", "Set your IANA timezone (like tz Asia/Dubai) and optionally set 24h"},
 		{"nick", nickCMD, "<name>", "Change your username"},
-		{"prompt", promptCMD, "<prompt>", "Change your promt. You can use \\w, \\h, and \\u PS1 tags"},
+		{"prompt", promptCMD, "<prompt>", "Change your promt. You can use \\w, \\W, \\h, and \\u PS1 tags"},
 		{"pronouns", pronounsCMD, "@user|pronouns", "Set your pronouns or get another user's"},
 		{"theme", themeCMD, "<theme>|list", "Change the syntax highlighting theme"},
 		{"rest", commandsRestCMD, "", "Uncommon commands list"}}
@@ -338,6 +338,7 @@ func bellCMD(rest string, u *User) {
 }
 
 func cdCMD(rest string, u *User) {
+	defer u.showPrompt()
 	if u.messaging != nil {
 		u.messaging = nil
 		u.writeln(Devbot, "Left private chat")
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -737,6 +739,41 @@ func (u *User) changeRoom(r *Room) {
 	u.room.broadcast("", Green.Paint(" --> ")+u.Name+" has joined "+Blue.Paint(u.room.name))
 }
 
+func (u *User) showPrompt() {
```
quackduck, line 742: is there a way to only go through the whole prompt generating function when the prompt changes? right now it generates your prompt string from scratch every single time

Arkaeriit: That's a good idea. I will try to implement it.

main.go
```diff
@@ -53,11 +53,13 @@ type Room struct {
...
 // Exported fields represent ones saved to disk. (see also: User.savePrefs())
 type User struct {
-	Name     string
-	Pronouns []string
-	Bio      string
-	session  ssh.Session
-	term     *terminal.Terminal
+	Name           string
+	Prompt         string
+	FormattedPrompt string
```
quackduck, line 58: wait why is this exported. does it need to be saved?

main.go
```diff
@@ -737,6 +741,45 @@ func (u *User) changeRoom(r *Room) {
...
+				u.FormattedPrompt += string(c)
+			}
+		} else {
+			u.FormattedPrompt += string(c)
+		}
+	}
+	u.showPrompt()
+}
+
+func (u *User) showPrompt() {
```
quackduck, line 779: I think we could just inline this function (delete it and replace all its usages by that one line of code inside this function). I can do that after accepting your PR though, so this doesn't block a merge.

commands.go
```diff
@@ -715,6 +723,11 @@ func manCMD(rest string, u *User) {
 		return
 	}
 
+	if rest == "prompt" {
+		u.room.broadcast(Devbot, "`prompt <prompt>` sets your prompt\\n\\nYou can use some bash PS1 tags in it.  \\nThe supported tags are:  \\n* \\u: your user name\\n* \\h, \\H: devzat colored like your username\\n* \\t, \\T: the time in your preferred formatting\\n* \\w: the current room\\n* \\W: the current room with #main aliased to ~\\n* \\S: a space character\\n\\nThe default prompt is `\\u:\\S`.")
```
quackduck, line 727: use backticks: "`" around this text so less escaping is needed. I don't think you need to have backticks around "prompt \<prompt\>"

main.go
Arkaeriit: It doesn't needs to be exported (and it should not be as it needs to be recomputed depending on the user's color) but it needs to saved alongside the user's data for the time the user is logged in.

If you know a better way to do so I would love to know it as the implementation I made is sub-optimal.

commands.go
Arkaeriit: Will do. Thank for this input.

main.go
Arkaeriit: Aren't you afraid this would impete readability. I know the Go compiler is not great at optimisation but this should not be too much of a problem.

main.go
quackduck: you can leave it unexported and it'll still be saved for the duration that the user's logged in: `formattedPrompt`

Arkaeriit: Great tip. Thank you.

main.go
Arkaeriit: Looking at the disassembly, this call is already inlined by the compiler.

```
		u.showPrompt()
  d54f72:	90                   	nop
	u.term.SetPrompt(u.formattedPrompt)
  d54f73:	48 8b 8c 24 38 01 00 	mov    0x138(%rsp),%rcx
  d54f7a:	00 
  d54f7b:	48 8b 51 68          	mov    0x68(%rcx),%rdx
  d54f7f:	48 8b 71 20          	mov    0x20(%rcx),%rsi
  d54f83:	48 8b 79 28          	mov    0x28(%rcx),%rdi
  d54f87:	48 89 d0             	mov    %rdx,%rax
  d54f8a:	48 89 f3             	mov    %rsi,%rbx
  d54f8d:	48 89 f9             	mov    %rdi,%rcx
  d54f90:	e8 ab 75 cf ff       	callq  a4c540 <github.com/quackduck/term.(*Terminal).SetPrompt>
```

I feel like it is a good idea to have the `showPrompt` in his own function if we want to change the behavior of the prompt in the future. (I heard someone wanted to add a black line if front of it or something)

What do you think about it?

quackduck: I wasn’t really worried about the compiler output. I just thought that the function wasn’t really enhancing readability much at all, but if we expect to make further prompt changes, this is not a bad idea. Let’s keep it.

main.go
```diff
@@ -737,6 +740,45 @@ func (u *User) changeRoom(r *Room) {
...
+	for _, c := range u.Prompt {
+		if c == '\\' {
+			last_escaped = true
+		} else if last_escaped {
+			last_escaped = false
+			switch c {
+			case 'u':
+				u.formattedPrompt += u.Name
+			case 'w':
+				u.formattedPrompt += copyColor(u.room.name, u.Name)
```
quackduck, line 755: I'm trying to think about whether a copyColor function is necessary. This could be replaced by a call to applyColorToData(<data>, u.Color, u.ColorBG). We wouldn't need a whole new function with possible new hard-to-track-down ANSI bugs. We've already used applyColorToData in all other such cases (such as when displaying pronouns).

Arkaeriit: As of now, there is no way to know the real color of the username based on the values of u.Color and u.ColorBG. Indeed such color scheme such as a list of color are using a random value when the name is colored. In cases such as with the rainbow color, the random value changes wildly the color rendering.

In such case, the color of the name and hostname can be very different, resulting in quite a ugly render.

Arkaeriit: One way to fix that would be to generate a rng seed before coloring a name and storing that seed in the user info. But I feel like this would be a bit of an overkill for our problem.

PR #165, opened 2023-06-08
Improve the sunset color scheme.

Arkaeriit:
The first letter was a bit too green.


Arkaeriit:
New: 
image

Old:
image

PR #166, opened 2023-06-09
Better install

Arkaeriit:
It is quite painful to self-host Devzat as the server crashes when it does not find some needed files such as the private key and the ban list.

This PR generates those files when needed.

### DIFF
--- a/util.go
+++ b/util.go
@@ -2,10 +2,8 @@ package main
 
 import (
 	"bytes"
-	trueRand "crypto/rand"
-	"crypto/rsa"
+	"crypto/ed25519"
 	"crypto/sha256"
-	"crypto/x509"
 	_ "embed"
 	"encoding/hex"
 	"encoding/json"
@@ -21,8 +19,11 @@ import (
 	"time"
 
 	"github.com/acarl005/stripansi"
+	"github.com/caarlos0/sshmarshal"
+	"github.com/fatih/color"
 	"github.com/gliderlabs/ssh"
 	markdown "github.com/quackduck/go-term-markdown"
+	cryptoSSH "golang.org/x/crypto/ssh"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


util.go
```diff
@@ -354,3 +369,22 @@ func fmtTime(u *User, lastStamp time.Time) string {
 	}
 	return lastStamp.In(u.Timezone.Location).Format("3:04")
 }
+
+// See if the public key is there and if it is not, try to create.
+func checkKey(keyPath string) {
```
quackduck, line 374: This function returns silently both when keys are already there and when there is a legitimate error. Those legitimate errors need to be communicated to the user. Also, if we do end up trying to generate keys, let the user know that keys were made and saved in [paths]

util.go
```diff
@@ -173,6 +177,17 @@ func readBans() {
 	if err != nil && !os.IsNotExist(err) { // if there is an error and it is not a "file does not exist" error
 		Log.Println(err)
 		return
+	} else if err != nil {
```
quackduck, line 180: We don't need to create the file. We should just modify the previous statement so that it doesn't print a file does not exist error but does make the function return silently. The file will be created when writing bans.

util.go
```diff
@@ -354,3 +369,22 @@ func fmtTime(u *User, lastStamp time.Time) string {
...
+		return
+	}
+
+	// generate and write private key as PEM
+	privateKeyFile, err := os.Create(keyPath)
+	if err != nil {
+		return
+	}
+	defer privateKeyFile.Close()
+	privateKeyPEM := &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(privateKey)}
```
quackduck, line 386: Do you think we should rely on RSA? Should we pick ECDSA instead? What do you think

quackduck (changes requested):

quackduck:
Good PR idea. This should make self hosting so painless

util.go
Arkaeriit: As the bans file should be written when using the ban command, it makes sense not to need it. Fair point.

util.go
Arkaeriit: I choose RSA because of how ubiquitous is it. RSA 4096 bits should be safe enough for some times. But if you would prefer to use a new and trendy elliptic curved based algorithm, lets do it.

I have no strong opinion one way or another, it's your call.

quackduck: LOL new and trendy. So true. Yes let’s go for the new and trendy elliptic curve stuff. That whole ssh-rsa-2 thing has made me saltier towards rsa.

Arkaeriit: Good point.

quackduck (approved):

PR #168, opened 2023-06-14
More UNIX commands

Arkaeriit:
This PR adds the `uname` and `uptime` UNIX command tweaked to show info relevant to Devzat.


quackduck:
Can we condense the build script into a single line? Readability doesn't suffer that much.

Arkaeriit:
> Can we condense the build script into a single line? Readability doesn't suffer that much.

What benefit would it bring? The readability will greatly suffer, it's already a bit painful to understand the quoting rules for the `go build` argument so making it a one liner would make it hellish. Furthermore, if we want to change the content of the message, it's easier to have the different parts laid out in multiple lines rather than to have subshells in subshell in subshell.

On an adjacent note, I arbitrarily chose the content of the messages outputted by the `uname` and `uptime` command. If you want them in a special formatting, please tell me.

Arkaeriit:
> Can we condense the build script into a single line? Readability doesn't suffer that much.

It would look like that:

```sh
echo go build -ldflags "-X 'main.unameMsg=$(TZ=UTC date "+Built from commit $(git log --pretty=format:%h --abbrev-commit | head -n 1) on the %G-%m-%d at %H:%M (UTC)")'"
```

So not very elegant.

quackduck:
What about now with all the formatting done in go? what does the one-liner look like now

Arkaeriit:
> What about now with all the formatting done in go? what does the one-liner look like now

```sh
tz=UTC go build -ldflags "-X 'main.unameCommit=$(git log --pretty=format:%h --abbrev-commit | head -n 1)' -X 'main.unameTz=$tz' -X 'main.unameDate=$(TZ=$tz date "+%G-%m-%d")' -X 'main.unameTime=$(TZ=$tz date "+%H:%M")'"
```

220 chars 😬

quackduck:
I think we can merge the date and time fields into a single variable. Also, `uname -a` for me does use the default date format, which is short and convenient. I've shortened the git command too.

99 chars:
```
go build -ldflags "-X 'main.unameCommit=$(git rev-parse --short HEAD)' -X 'main.unameTime=$(date)'"
```

Arkaeriit:
> I think we can merge the date and time fields into a single variable. Also, `uname -a` for me does use the default date format, which is short and convenient. I've shortened the git command too.
> 
> 99 chars:
> 
> ```
> go build -ldflags "-X 'main.unameCommit=$(git rev-parse --short HEAD)' -X 'main.unameTime=$(date)'"
> ```

This does not works very well. Indeed, using the default behavior of dote meas that the command will use the system's locale for the date formatting which might not be good (on my system, I use an US locale for language but a French one for everything else). Furthermore, it will use the system's timezone which might not be suitable for a service used all over the world. Here is what it looks when I build Devzat:
```
max: uname
Devzat c9f0a7e: mar. 27 juin 2023 11:09:14 CEST
```
Pretty bad, I think.

Arkaeriit:
Good idea on changing the formatting of the uptime command, the one I made was a bit quirky. I made a new pass on it to make it look a lot like the default `uptime`. What do you think of it? I will run it on `bobignou.red -p 2222` if you want to test it after a day of run.

quackduck:
This is what uname looks like for me:
```
Devzat c9f0a7e: Tue Jun 27 12:12:11 PDT 2023
```

I think exposing build locale is okay. For example, here's the uname output on my mac:

```
Darwin ishans-air.lan 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:40:17 PDT 2023; root:xnu-8796.121.2~3/RELEASE_ARM64_T8112 arm64
```

I wonder if maybe a long hash would look nicer. Maybe we should remove that colon.

commands.go
```diff
@@ -805,6 +807,20 @@ func unameCMD(rest string, u *User) {
 	u.room.broadcast("", "Devzat "+unameCommit+": "+unameTime)
 }
 
+func floor(f float64) int {
```
quackduck, line 810: `int(math.Floor(f))` is equivalent to `int(f)`

quackduck: i've added some simplifications

Arkaeriit:
> I wonder if maybe a long hash would look nicer. Maybe we should remove that colon.

I feel like it would be prettier with the long hash and the colon.

On my system, it looks like that:
```
Linux arrakis 6.3.9-1-default #1 SMP PREEMPT_DYNAMIC Thu Jun 22 03:53:43 UTC 2023 (0df701d) x86_64 x86_64 x86_64 GNU/Linux
```
It's fun to see how much different it is.

quackduck:
> I feel like it would be prettier with the long hash and the colon.

okay sure let’s change the format to this

quackduck:
Thoughts on this one?

Arkaeriit:
> Thoughts on this one?

I rather like it.

PR #172, opened 2023-07-02
Updated censor.go to work with the improved flexibility of falsepositives.go

CaenJones:
I changed the `detector` variable to use the new `goaway.DefaultFalsePositives` function as the false positives filter. This lets any word to pass without triggering the default profanities.

The code for the edited falsepositives.go is at https://github.com/CaenJones/go-away/blob/master/falsepositives.go

Thank you to:
QuackDuck
Arkaeriit
Google
Derplord264

*this will also adapt the admin rights to my home pc not my school laptop :)

### DIFF
--- a/censor.go
+++ b/censor.go
@@ -1,37 +1,29 @@
 package main
 
 import (
-	"fmt"
-	"strings"
-
-	"github.com/CaenJones/goaway"
+	"encoding/base64"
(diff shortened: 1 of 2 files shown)


quackduck:
Try to develop with a clone of Devzat so you can make sure compile works

PR #177, opened 2023-07-10
Updated colors.go for easier usage + moving admin account to personal computer

CaenJones:
Updated the color package import to github.com/fatih/color for simpler color formatting.

Modified the color definitions using the new package's color structure

Modified the applyStyles function to apply colors using the new color package.

Updated the getStyle function to use the new color definitions.

Modified the changeColor method of the User struct to utilize the new color package.

CMD needed: go import github.com/fatih/color

Thank you to derplord256 for your help

### DIFF
--- a/colors.go
+++ b/colors.go
@@ -10,57 +10,57 @@ import (
 
 	"github.com/acarl005/stripansi"
 	chromastyles "github.com/alecthomas/chroma/styles"
-	"github.com/jwalton/gchalk"
+	"github.com/fatih/color"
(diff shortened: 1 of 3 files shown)


quackduck:
We actually used to use fatih/color before that current package. We switched for more granular control of color. You'll see that fatih/color does not allow for truecolor (proper 24 bit rgb color) support. This PR reverts a lot of progress :(

quackduck:
You can make the ID update a separate PR

CaenJones:
ok :(

PR #180, opened 2023-07-17
Add recommended specs to admins manual

CaenJones:
It is what it says

### DIFF
--- a/Admin's Manual.md
+++ b/Admin's Manual.md
@@ -4,6 +4,15 @@ This document is for those who want to manage a self-hosted Devzat server.
 
 Feel free to make a [new issue](https://github.com/quackduck/devzat/issues) if something doesn't work.
 
+
+## Recommended Specs
+  * 1 Core
+  * 0.5 GB RAM
+  * 20 GB Storage
+  * 50 MBPS network speed or higher
+
+ Any Linux-based OS will work well with Devzat. We recommend ubuntu 22.04 LTS (as of 2023).
+ 
 ## Installation
 ```shell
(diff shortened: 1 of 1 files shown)


quackduck:
How did you come up with these specs?

CaenJones:
I ran devzat in a VM at home. This was the area where it preformed well with the least processing power

quackduck:
I'm pretty sure the main Devzat instance has an internet speed available to it measured in kbps

CaenJones:
what would you recommend it being then?

quackduck:
Usually people only add in recommended specs when some sort of quite heavy application is being run. Devzat is text based (thus not needing much network bandwidth) and uses barely any memory or storage.

CaenJones:
it would make it so they do not over-allocate resources for the hosting

quackduck:
Yeah we could add in a lil note about how devzat is so lightweight you could even run it on a base level raspberry pico or something.

CaenJones:
ok then, one sec...

quackduck:
"20 GB Storage" makes it seem like Devzat itself will take up at least 20 gigs of storage

CaenJones:
ok, I changed it

quackduck:
Currently the data storage on the main server is 15 mb for the executable + 5 mb for user data

quackduck:
Let me just confirm it will actually run on a pico. Asking around...

CaenJones:
works with pi zero, will post extensive guide soon :P

CaenJones:
wat

quackduck:
this is my reaction too

PR #182, opened 2023-07-19
Fix Discord Bridge CD Issue

CaenJones:
This  fixes the issue with Discord  where using cd which caused the user's Discord account name to be displayed along with the devzat username when typing in cd" message. The problem was   the discordMessageHandler function, which has been changed  to skip processing when the message content is cd.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -6,11 +6,14 @@ import (
 	_ "image/jpeg"
 	_ "image/png"
 	"math"
+	"errors"
+	"sort"
 	"math/rand"
 	"sort"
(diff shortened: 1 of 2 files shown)


quackduck:
I want to do some thinking about whether this is the right way to solve this problem. Maybe this cd message should just never be sent. But that would require sending the messages to one bridge but not the other. I'm also curious about @Arkaeriit's thoughts on this.
image

CaenJones:
Ok

quackduck:
Why did you close this?

PR #183, opened 2023-07-19
devbot improvements

TAG-Epic:
- shows devbot on ls list

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -787,7 +787,6 @@ func lsCMD(rest string, u *User) {
 		for _, us := range u.room.users {
 			s += us.id + " " + us.Name + "  \n"
 		}
-        s += strings.Repeat("0", 64) + Devbot + " \n"
 		u.room.broadcast("", s)
 		return
 	}
--- a/commands.go
+++ b/commands.go
@@ -534,7 +534,7 @@ func banCMD(line string, u *User) {
 
     boreName := "http://bore.pub:1337"
     if split[0] == boreName {
-        bore, boreOnline = findUserByName(u.room, boreName)
+        bore, boreOnline := findUserByName(u.room, boreName)
         if boreOnline {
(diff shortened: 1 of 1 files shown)


quackduck:
~~this one is easy to review~~

quackduck:
“Oops”

commands.go
```diff
@@ -531,6 +531,16 @@ func banCMD(line string, u *User) {
 	var victim *User
 	var ok bool
 	banner := u.Name
+
+    boreName := "http://bore.pub:1337"
+    if split[0] == boreName {
+        bore, boreOnline := findUserByName(u.room, boreName)
+        if boreOnline {
+            u.room.broadcast(bore.Name, "Nice try devbot.")
+            u.room.broadcast("", "devbot has been banned by " + banner)
```
quackduck, line 540: You do know this won't actually disable devbot right.

quackduck:
What's the rationale for adding devbot to the user list?

commands.go
TAG-Epic: Yes

quackduck:
Also, I don't think it would be good idea for me to introduce this special case for _all_ instances of Devzat. I might allow plugins that treat a username in a special way to run on the main server though.

TAG-Epic:
> What's the rationale for adding devbot to the user list?

Devbot is always here, he should be considered a user

commands.go
```diff
@@ -777,6 +787,7 @@ func lsCMD(rest string, u *User) {
 		for _, us := range u.room.users {
 			s += us.id + " " + us.Name + "  \n"
 		}
+        s += strings.Repeat("0", 64) + Devbot + " \n"
```
quackduck, line 790: Adding devbot to the ID list isn't useful.
```suggestion
```

quackduck:
Okay let's try it

PR #186, opened 2023-07-22
Fix Discord cd issue & Add mute/unmute function to devzat

CaenJones:
I modified the discordMessageHandler function, which has been changed  to skip processing when the message content is cd or cd <dir>

I have also added

/mute <user> - Toggles muting for the specified <user>. When a user is muted, their messages will not be broadcasted to the chat room.

/unmute <user> - Toggles unmuting for the specified <user>. When a user is unmuted, their messages will be broadcasted to the chat room as usual.

I hope this works :P
-Caen

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -8,7 +8,6 @@ import (
 	"math"
 	"math/rand"
 	"sort"
-	"errors"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


quackduck:
None of the types used by `github.com/shazow/ssh-chat` are compatible with devzat.

PR #188, opened 2023-07-24
Add Magic-8-Ball function

CaenJones:
It is fun!

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -923,5 +923,6 @@ func getMagic8BallResponse() string {
 }
 
 func returneightball() {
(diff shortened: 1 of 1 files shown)


CaenJones:
Why it not working?!?!?!?!?!

PPTide:
You are printing to the console not actually to devzat xD

PPTide:
U can use `u.room.broadcast("8-ball", response)`

quackduck:
This would make for an excellent plugin!

CaenJones:
Thanks!

PR #189, opened 2023-07-24
Fix Discord Bridge CD Issue

CaenJones:
This fixes the issue with Discord where using cd which caused the user's Discord account name to be displayed along with the devzat username when typing in cd" message. The problem was the discordMessageHandler function, which has been changed to skip processing when the message content is cd


quackduck:
Don't close PRs if I don't ask you to!

PR #191, opened 2023-08-01
Devzat status website

CaenJones:
I have created a devzat status site using uptimerobot. Status site will be shown when a new user joins.


quackduck:
In order of importance:

1. This would only report status for the main server, but print the same message even on the other servers.
2. I just killed devzat and waited for the site to refresh to check if it works. It still reports devzat as up even though it's down.
3. Checking if devzat is down is as simple as trying to login and seeing if it works. I don't see a need for a whole website that reports this information.

Please make sure a need for a PR exists first by making a github issue for it or by chatting about it with me and others (eg. @Arkaeriit) on devzat.

PR #195, opened 2023-08-02
Add Magic8Ball Function

CaenJones:
This will create a function so when a user types 8ball (input) it will generate a response from a predefined list of answers.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -878,7 +878,7 @@ func EightBallCMD(_ string, u *User) {
 	response := getMagic8BallResponse()
 
 	// Print the response
-	u.room.broadcast("8-Ball", response)
+	u.room.broadcast("8ball", response)
 }
 
 func getMagic8BallResponse() string {
--- a/commands.go
+++ b/commands.go
@@ -876,26 +876,13 @@ func neofetchCMD(_ string, u *User) {
 
 func EightBallCMD(_ string, u *User) {
 	responses := []string{
-		"It is certain, ",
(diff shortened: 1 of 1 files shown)


commands.go
```diff
@@ -872,3 +873,40 @@ func neofetchCMD(_ string, u *User) {
...
+		"My sources say no.",
+		"Outlook not so good.",
+		"Very doubtful.",
+	}
+
+	rand.Seed(time.Now().UnixNano())
+	i := rand.Intn(len(responses))
+
+	return responses[i]
+}
```
quackduck, line 912:
```suggestion
func EightBallCMD(_ string, u *User) {
	responses := []string{
		"It is certain, ",
		"It is decidedly so, ",
		"Without a doubt, ",
		"Yes, definitely, ",
		"You may rely on it, ",
		"As I see it, yes, ",
		"Most likely, ",
		"Outlook good, ",
		"Yes, ",
		"Signs point to yes, ",
		"Reply hazy, try again, ",
		"Ask again later, ",
		"Better not tell you now, ",
		"Cannot predict now, ",
		"Concentrate and ask again, ",
		"Don't count on it, ",
		"My reply is no, ",
		"My sources say no, ",
		"Outlook not so good, ",
		"Very doubtful, ",
	}
	u.room.broadcast("8ball", responses[rand.Intn(len(responses))] + u.Name)
}
```

quackduck:
Looks good!

PR #196, opened 2023-08-02
Fix lstokens command

Arkaeriit:
There is two issues with the current implementation of `lstoken`.
* The first token is numbered '2'.
* The numbers are not 0 padded which cause the column to be misaligned which hurts the Feng Shui of the output.

miss_lstoken

This PR fixes those two issues.

good_lstoken

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"math"
 	"net"
 	"os"
 	"regexp"
--- a/rpc.go
+++ b/rpc.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"os"
 	"regexp"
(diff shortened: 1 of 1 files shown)


rpc.go
```diff
@@ -374,9 +375,11 @@ func lsTokensCMD(_ string, u *User) {
 	}
 	msg := "Tokens:  \n"
 	i := 0
+	tokenNumberSize := int(math.Log10(float64(len(Tokens)))) + 1
```
quackduck, line 378:
```suggestion
	tokenNumberSize := len(fmt.Sprint(len(Tokens)))
```
Would this work too?

rpc.go
```diff
@@ -374,9 +375,11 @@ func lsTokensCMD(_ string, u *User) {
...
+	tokenNumberFormatFlag := "%0" + strconv.Itoa(tokenNumberSize) + "d"
```
quackduck, line 379:
```suggestion
	tokenNumberFormatFlag := "%0" + fmt.Sprint(len(Tokens)) + "d"
```

rpc.go
```diff
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
```
quackduck, line 10:
```suggestion
```

PR #197, opened 2023-08-02
Center logo in README.MD

CaenJones:


quackduck:
You're cutting a lot of documentation from the page. Your PR says this is a reformat, but it actually edits a lot of the actual content. I like the idea of centering the logo.

CaenJones:
True, but most of the documentation is just copied from the chats help menus or available on the admins or integration manuals. They do not need to be included on the README and add unnecessary clutter.

CaenJones:
But I did bring back some things that might be useful

quackduck:
I'm sorry, I don't really see this PR adding much in its current state.
image

PR #203, opened 2023-08-10
Add Ban_Reason to ban function

CaenJones:
the new format will display the username + reason for being banned


quackduck:
thanks!

PR #204, opened 2023-08-10
Mention WolfRam Alpha in README + Admin's Manual

CaenJones:


Arkaeriit:
As the wolf commands comes from a plugin, it doesn't feel wise to mention it here. If someone wants to self-host Devzat, they won't have that command available and they might be confused.

If we want to acknowledge the plugins, we might have a list of all known plugins somewhere, but I don't think we should do more.

CaenJones:
I see

quackduck:
We should definitely have a list of plugins maybe near the bottom of the readme. thoughts, @Arkaeriit?

Arkaeriit:
> We should definitely have a list of plugins maybe near the bottom of the readme. thoughts, @Arkaeriit?

I think you're right. I will open an issue so that we can start to curate a list.

PR #205, opened 2023-08-11
Move the invocation of middleware

Arkaeriit:
The current use of middleware had two issue:
* As the message was processed before we check if the message is a DM, middleware admin could snoop on any DM. This is not great for privacy.
* Bridge messages didn't go though middleware. On a first thought, this might have been seen as a good thing considering this reduce the difference between what is on the bridge and what is on the SSH interface. But there is already a lot of difference between messages sent on bridges and how the appear on the SSH interface. Messages can be edited on bridges, custom emotes are used there, and go-away censor the messages from bridges.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -115,6 +115,7 @@ func runCommands(line string, u *User) {
 	}
 
 	// Now we know it is not a DM, so this is a safe place to add the hook for sending the event to plugins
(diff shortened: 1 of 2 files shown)


quackduck:
Thank you, your prs always make sense

PR #207, opened 2023-08-17
Add Dialogue to PromptCMD

CaenJones:
When I attempted to use "prompt" in the past, it hid my username so I could only see it after I sent a message.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -492,7 +492,7 @@ func promptCMD(line string, u *User) {
     u.formatPrompt()
 
     if line == "" {
-        User.broadcast(Devbot, "Your prompt is now empty. Did you mean to get more info about your prompt? Run man prompt for more info")
+        MainRoom.broadcast(Devbot, "Your prompt is now empty. Did you mean to get more info about your prompt? Run man prompt for more info")
     }
 }
 
--- a/commands.go
+++ b/commands.go
@@ -492,8 +492,8 @@ func promptCMD(line string, u *User) {
     u.formatPrompt()
 
     if line == "" {
-	User.broadcast(Devbot, ("Your prompt is now empty. Did you mean to get more info about your prompt? Run man prompt for more info")
-	          }
+        User.broadcast(Devbot, "Your prompt is now empty. Did you mean to get more info about your prompt? Run man prompt for more info")
+    }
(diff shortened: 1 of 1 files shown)


CaenJones:
YEAH!

quackduck:
Afaict, the change here is that empty prompts aren’t allowed and print an error message. I think empty prompts should be allowed but we can have devbot DM something like “(Your prompt is now empty. Did you mean to get more info about your prompt? Run `man prompt` for more info)”. Also we don’t really need a whole new setPrompt function: just change the original one to print a message! We dont need the errors package either :)

commands.go
```diff
@@ -488,8 +488,14 @@ func nickCMD(line string, u *User) {
 }
 
 func promptCMD(line string, u *User) {
-	u.Prompt = line
-	u.formatPrompt()
+    u.Prompt = line
+    u.formatPrompt()
+
+    if line == "" {
+        fmt.Println("(Your prompt is now empty. Did you mean to get more info about your prompt? Run man prompt for more info)")
```
quackduck, line 495: You want to write to the user, not stdout!

commands.go
```diff
@@ -488,8 +488,14 @@ func nickCMD(line string, u *User) {
...
+    } else {
```
quackduck, line 496: I don't think we need a success message

quackduck (changes requested):

commands.go
```diff
@@ -488,8 +488,12 @@ func nickCMD(line string, u *User) {
...
 
 func promptCMD(line string, u *User) {
-	u.Prompt = line
-	u.formatPrompt()
+    u.Prompt = line
+    u.formatPrompt()
+
+    if line == "" {
+        MainRoom.broadcast(Devbot, "Your prompt is now empty. Did you mean to get more info about your prompt? Run man prompt for more info")
+    }
```
quackduck, line 496:
```suggestion
	u.Prompt = line
	u.formatPrompt()
	if line == "" {
		u.writeln(Devbot, "(Your prompt is now empty. Did you mean to get more info about your prompt? Run `man prompt` for more info)")
	}
```

quackduck (approved):

PR #209, opened 2023-09-26
Prevent FalsePostives in Censor.go (for real this time!)

CaenJones:
by adding .WithWholeWordMatch(true) the detector will only block bad words if there is a complete match.
( when your attempt to do a rewrite of falsepositives.go on go-away was not needed :P )


CaenJones:
wait, what?

PR #211, opened 2023-11-08
Devzat logger CSV File Integration  

CaenJones:
This records what time a user joins and leaves Devzat onto a csv file. This file will not be erased unlike the normal Devzat logfile.


main.go
CaenJones, line 1: for the error on line 909, because logging is disabled by default, `(untyped bool constant) is not used` is not an issue

main.go
```diff
@@ -895,11 +896,49 @@ func calculateLinesTaken(u *User, s string, width int) {
...
 	for i := 0; i < len(b); i++ {
 		if b[i].Addr == addr || b[i].ID == id {
 			return true
 		}
 	}
 	return false
 }
+
+func logUserActivity(user *User, action string, loggingEnabled bool) {	
+    if !loggingEnabled { false //logging on Devzat is disabled by default, to enable it please switch the statement from false to true
```
CaenJones, line 909: for the error on line 909, because logging is disabled by default, (untyped bool constant) is not used is not an issue

quackduck:
Your code needs to compile!

Arkaeriit:
> Your code needs to compile!

You see a code that can't compile, I see a code with 0 runtime errors.

quackduck:
So true

quackduck:
Why did you close the PR?

PR #217, opened 2024-02-01
Fix `man prompt`

Arkaeriit:
The escaping was made for the old markdown engine
and looked dorky with the new one.


Arkaeriit:
The build fails but #216 fixes it.

PR #223, opened 2024-07-20
Add very basic docker support

### LINKED ISSUE #222: Add docker support
I think that adding docker support will be useful for a lot of users.

While shipping images on dockerhub with releases is ideal, we can still have a basic integration with mention in the docs & a dockerfile in the repo.

peauc:
Fix #222

Add basic docker support for self hosted build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM golang:1.22.5-alpine
+
+WORKDIR /devzat
+
+COPY . .
+
+RUN go build
+
+EXPOSE 8080
+
+CMD ./devzat
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+.github
+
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.22.5-alpine
+
+WORKDIR /devzat
+
+COPY . .
+
+RUN go build
+
+EXPOSE 8080
```
Aeron, line 9: The default port for the server is `2221`, and the alternative one is `8080`. Is there any specific reason why only the latter is specified? The `EXPOSE` instruction is only informative but still.

peauc: During my tests i've found that I can connect to the chatroom using port 8080 so I replicated that in the dockerfile. I've spent no more than 15 minutes on the dockerfile + PR, no idea if other configurations works but interested in figuring it out

quackduck: there's a bunch of ports used. you'd probably want to expose the rpc one and also both the ssh ones. please have a look at the source

Arkaeriit:
Devzat uses the file system to store ssh keys, server configuration, and user configuration. All those should be permanent after a Devzat restart. I am not a Docker expert but it looks like the files are erased after closing the docker container.

quackduck:
What's the relationship here with the other docker pr?

peauc:
@quackduck there is none, im closing this one

PR #230, opened 2024-08-09
Better bans

Arkaeriit:
This PR adds the following improvement related to bans:
* Temporary bans are robust against server restart.
* User banned temporary can see how long they have to wait.
* Cleaner in-code ban API.
* Devbot banning and anti-spam banning are respectively one minute and one month instead of being forever.

### DIFF
--- a/main.go
+++ b/main.go
@@ -439,7 +439,7 @@ func newUser(s ssh.Session) *User {
 			u.writeln(Devbot, "**You are banned**. If you feel this was a mistake, please reach out to the server admin. Include the following information: [ID "+u.id+"]")
 			if banInfo.UseTime {
 				when := time.Until(banInfo.UnbanTime)
-				u.writeln(Devbot, "You will be unbaned in "+formatDuration(when)+".")
+				u.writeln(Devbot, "You will be unbaned in "+printPrettyDuration(when)+".")
 			}
 			s.Close()
 			return nil
--- a/main.go
+++ b/main.go
@@ -463,7 +463,7 @@ func newUser(s ssh.Session) *User {
 		IDandIPsToTimesJoinedInMin[u.id]--
 	})
 	if IDandIPsToTimesJoinedInMin[u.addr] > 6 || IDandIPsToTimesJoinedInMin[u.id] > 6 {
-		u.ban("", false, time.Now())
+		u.banForever("")
 		MainRoom.broadcast(Devbot, u.Name+" has been banned automatically. ID: "+u.id)
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -410,13 +412,28 @@ func newUser(s ssh.Session) *User {
...
 
+	banInfo := getBan(Bans, u.addr, u.id)
+	if banInfo != nil {
+		if banInfo.UseTime && banInfo.UnbanTime.Before(time.Now()) {
+			unbanIDorIP(banInfo.ID)
+		} else {
+			Log.Println("Rejected " + u.Name + " [" + host + "] (banned)")
+			u.writeln(Devbot, "**You are banned**. If you feel this was a mistake, please reach out to the server admin. Include the following information: [ID "+u.id+"]")
+			if banInfo.UseTime {
+				u.writeln(Devbot, "You will be unbaned on "+banInfo.UnbanTime.Format(time.RFC3339))
```
quackduck, line 430: use a duration here instead of an absolute time

Arkaeriit: It looks like that:
```
devbot: You will be unbaned in 41 days, 15 hours, 56 minutes, and 8 seconds.
```
or like that:
```
devbot: You will be unbaned in 13 seconds.
```

quackduck: It's quite verbose. Which could be fine. But we do also have a pretty duration printer here: https://github.com/Arkaeriit/devzat/blob/c98c58e03fd3703f5d500424e886e5f06fe0768d/util.go#L420

what do you think?

main.go
```diff
@@ -43,8 +43,10 @@ const (
 )
 
 type Ban struct {
```
quackduck, line 45: Is this backward compatible?

main.go
Arkaeriit: Ho right, I didn't remember this function. `printPrettyDuration` is not very suitable for duration longer than a day, so not great for bans.
But maybe we can merge `printPrettyDuration` and `formatDuration` to prevent code duplication.
I'll make a proposition of it.

main.go
Arkaeriit: Yes, old bans are converted to new bans automatically (they default to being indefinite) and new bans are converted to old ones.

main.go
Arkaeriit: I went with full unit names. So it changes the way the relative time display is. If you don't like it, I will go back to units symbols.
image

util.go
```diff
@@ -418,12 +418,26 @@ func holidaysCheck(u *User) {
 }
 
 func printPrettyDuration(d time.Duration) string {
```
quackduck, line 420: This makes the relative timestamps too long, I prefer the older format

main.go
quackduck: wdym by the new bans are converted to old ones?

main.go
quackduck: I see what you mean. maybe it's better to keep it short but add ability to display days using the unit 'd'

util.go
Arkaeriit: Understood, I will revert it back.

main.go
Arkaeriit: A database generated in the old format can be read by a new Devzat, and a database generated in the new format can be read by an old Devzat.

quackduck: oh awesome!

main.go
Arkaeriit: I reverted back to using 1 letter units with no space between the numeric value and the unit.

Arkaeriit: It looks like that now: 
image

Arkaeriit:
I added an additional feature of storing names in the DB and displaying them when doing `lsbans`. This should make it easier to clean erroneous bans.

quackduck:
Okay I will accept this. I did have some questions about whether it's really necessary to store a bool when we could maybe have a special Time value indicating it's invalid (eg. 0). Also wondering whether we could merge the internals of temp ban and forever ban by just having an extremely large unban time (max time?). The question here isn't about whether these implementations are possible but whether they are good design.

I can accept for now and we can address these later.

quackduck:
Thank you very much Arkaeriit, especially for your insane patience. Took me really long..

PR #232, opened 2024-08-10
Join Docker support

0Knot:
Developed Dockerfile to support Docker Go compilation and export images

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,39 @@
+# Use the official Alpine-based Golang image as the build stage
+FROM golang:1.18-alpine AS builder
+
+# Set the working directory
+WORKDIR /go/src/github.com/quackduck/devzat
+
+# Copy the project source code into the container
+COPY . .
+
+# Install project dependencies
+RUN go mod download
+
+# Build the project
+RUN go build -o devzat .
+
+
+
+
+# Use the latest Alpine Linux image as the base for the final image
+FROM alpine:latest
+
+# Install necessary tools
+RUN apk add --no-cache openssh-server
+
+# Copy the compiled binary to the final image
+COPY --from=builder /go/src/github.com/quackduck/devzat /usr/local/bin/devzat
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+FROM alpine:latest
+
+# Install necessary tools
+RUN apk add --no-cache openssh-server
+
+# Copy the compiled binary to the final image
+COPY --from=builder /go/src/github.com/quackduck/devzat /usr/local/bin/devzat
+
+# Generate the SSH host key
+RUN ssh-keygen -qN '' -f /etc/ssh/ssh_host_rsa_key
```
PPTide, line 29: If i understand Docker correctly this would mean for a published Docker-Image everyone would have the same key. The key should either be generated by Devzat on startup or by the user themselves

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Expose the SSH port
+EXPOSE 22
```
PPTide, line 32: There are more ports used, please look at the source

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Configure and start the SSH service
+RUN sed -i 's/#PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config
+CMD ["/usr/sbin/sshd", "-D"]
+
+# Run the application (This line is commented out because the main service is the SSH server)
+# CMD ["devzat"]
```
PPTide, line 39: This is just wrong. Devzat runs as it's own ssh server. You should only be running devzat. (Please test before creating a PR this just runs a new ssh server)

PPTide (changes requested):

Dockerfile
0Knot: I'm very sorry, I submitted the Dockerfile for dev test, I will correct it and submit it again

PR #239, opened 2024-08-16
Rework Discord Webhooks with pictures

PPTide:
## Create Multiple Webhooks, one for each color
This is so that we don't have to edit Webhooks all the time and thus not run into a rate-limit (as often).

## Don't let discordgo retry on rate limit
Instead just send the message through the compact-mode

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -5,19 +5,17 @@ import (
 	"crypto/sha1"
 	"encoding/base64"
 	"encoding/hex"
+	"github.com/acarl005/stripansi"
+	"github.com/bwmarrin/discordgo"
+	"github.com/leaanthony/go-ansi-parser"
+	"github.com/quackduck/term"
+	"golang.org/x/image/draw"
 	"image"
 	"image/color"
 	"image/png"
 	"os"
 	"strconv"
 	"strings"
-	"time"
-
-	"github.com/acarl005/stripansi"
-	"github.com/bwmarrin/discordgo"
-	"github.com/leaanthony/go-ansi-parser"
-	"github.com/quackduck/term"
-	"golang.org/x/image/draw"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


PPTide:
I still want to clean-up the code a bit before it is ready to merge but i would love some feedback.

Arkaeriit:
Some remarks:
- There is quite a lot of magic numbers like "20" or "13". You should probably define them as const somewhere.
- You create a webhook list but you keep the image cache. Maybe you should merge the two together.

PPTide:
Merging the Webhook list and image cache would prob be possible but tedious and not worth it imo.

The resulting list would need to account for possible many to one entries as e.g. two users using `color blue` would share the same webhook/have the same image or a user using a solid color changing username can keep the same webhook. 

The webhooks should also be kept in a list that can be recovered from discord so that we don't need to delete all webhooks on startup.

quackduck:
The code is hard to read. Needs some refactoring. Also the `continue nextMsg` feels like a code smell.

discord.go
```diff
@@ -153,13 +190,22 @@ func discordMessageHandler(_ *discordgo.Session, m *discordgo.MessageCreate) {
...
+func (i *imgCache) add(user, image string) {
+	if len(*i) >= cacheSize {
+		// remove the first value
+		for k := range *i {
+			delete(*i, k)
+			break
+		}
+	}
+	(*i)[user] = image
+}
```
quackduck, line 208: I don't think you need to delete the previous value in a map. You can set it directly.

quackduck:
Creating an image is very resource intensive. So eg. we cannot be calculating the hash of the avatar every message.

discord.go
PPTide: i am trying to keep the map from getting to large, as the images themselves are stored in the map and users can easily change color or name and thus create extra entries in the cache. For that reason i believe there should be a limit as to not allow the ram usage to increase that much.

PPTide:
> Creating an image is very resource intensive. So eg. we cannot be calculating the hash of the avatar every message.

There is a cache for those images, so it is just a look up into a map after the user has sent the first message

discord.go
PPTide: Having the cache linked to a list of online members would be better but would also mean deeper changes into devzat's code that is not related to the discord bridge.

quackduck: ah nvm I see what the code does. we should make this an LRU cache.

discord.go
```diff
@@ -50,80 +50,104 @@ func discordInit() {
...
+					for _, wh := range webhooks {
+						found := false
+						for _, user := range users {
+							if wh.Name == shasum(createDiscordImage(user)) { // generating all the images is cashed
+								found = true
+								break
+							}
+						}
+						if !found {
+							err = sess.WebhookDelete(wh.ID, discordgo.WithRetryOnRatelimit(false))
```
quackduck, line 109: wh will always be the last item in the webhooks list, right? is this intentional? also if this is the intended behavior, you no longer need a loop to find the index of this webhook since you know it's the last one.

discord.go
```diff
@@ -50,80 +50,104 @@ func discordInit() {
...
+								sendDiscordCompactMessage(msg, txt, sess)
+								continue nextMsg
+							}
+							webhookIndex := -1
+							for i, w := range webhooks {
+								if w.ID == wh.ID {
+									webhookIndex = i
+									break
+								}
+							}
```
quackduck, line 121: You can skip this by using an index variable on line 100.

discord.go
```diff
@@ -50,80 +50,97 @@ func discordInit() {
...
+				// (discord can only have 15 webhooks per channel)
+				if webhook == nil && len(webhooks) > maxWebhookCount {
+					// generate a list of all users in all rooms
+					users := make([]string, 0, maxWebhookCount)
+					for _, room := range Rooms {
+						for _, user := range room.users {
+							users = append(users, user.Name)
+						}
+					}
+					users = append(users, "", Devbot)
```
quackduck, line 91: Doesn't feel right that this list is generated every message. Can we instead just update this list whenever users are added or removed?

quackduck: Also, what happens when plugins send messages?

quackduck: We could have one webhook that handles all miscellaneous situations

PPTide: > Doesn't feel right that this list is generated every message. Can we instead just update this list whenever users are added or removed?

This would need deeper changes outside of the discord bridge... i can try adding this in an expandable style or just for discord 🤷‍♀️

PPTide: > We could have one webhook that handles all miscellaneous situations

i could use the empy name webhook that is used for system commands, or add a difference between the max number of webhooks and the max number of users to use webhooks for so that webhooks are created for plugins, they might just get deleted after a user changes color or another plugin/the same plugin sends a message with different colors...

PR #244, opened 2024-09-20
`clear` command is not displayed to other users.

Arkaeriit:
When an user uses "clear" to refresh their screens, it's not displayed to other users. This is to reduce the noise during discussions.


CaenJones:
lmfao that quick? :P

Arkaeriit:
Merging as I can't stand Caen spamming `clear` anymore.

vcc3v:
Come on @Arkaeriit its not thaatttt bad :P

Issue #246, opened 2024-12-12
Crash when trying to fetch torlist


TLINDEN:
I want to operate devzat in an isolated environment. There's no dns available.

On startup it tries to fetch data from https://www.dan.me.uk/torlist/?exit which fails due to a dns error. However, devzat doesn't handle this and crashes instead:

```default
# DEVZAT_CONFIG=/usr/local/etc/devzat.yml daemon -P /tmp/devzat.pid -u nobody /usr/local/sbin/devzat
# Config loaded from /usr/local/etc/devzat.yml
2024/12/12 18:27:29 main.go:154: Get "https://www.dan.me.uk/torlist/?exit": dial tcp: lookup www.dan.me.uk on [::1]:53: read udp [::1]:39505->[::1]:53: read: connection refused
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x100086a]

goroutine 1 [running]:
main.main()
        /tmp/devzat/main.go:156 +0x1ca
```

The program should put out a warning that the torlist is unavailable or provide a config option to disable the feature.

Arkaeriit:
That's a very good point, thanks for noticing it.

Arkaeriit:
I've fixed it in #247, can you confirm that it works on your side?

TLINDEN:
@Arkaeriit Thanks a lot for fast response and fix! I can confirm, that your patch works as intended. Loving it :)

quackduck:
thank you arkaeriit!

PR #247, opened 2024-12-13
Recover from not being able to read TOR IPs.

Arkaeriit:
The error handling for when not being able to get the TOR IP list was broken, resulting in a crash as shown in #246.


Arkaeriit:
@quackduck as this issue prevents an user for using Devzat, I'll merge it. Feel free to revert it if it's not satisfactory for you.

Issue #248, opened 2025-01-06
add additional information to plugins


PPTide:
the api currently doesn't show system messages and user color (and probably other things) that would be useful especially for (for example) bridge-plugins.

another thing that would probably be useful are infos for online users.

i might want to work on something like this but want some feedback on the idea first.

quackduck:
good idea

PR #249, opened 2025-01-13
Update dependencies

PPTide:
This is needed for #248

I'm only updating the packages needed for a proto update


Arkaeriit:
Looks good to me. If it is still mergable I can merge it.

It looks like those changes are copied in your other MR. Is that right?

PR #251, opened 2025-01-28
Rpc add options for bridge

PPTide:
This creates a backwards compatible rpc update that enables plugins to get user color and system messages. This is usefully especially for bridges and similar


main.go
```diff
@@ -205,10 +205,18 @@ func main() {
...
 
-func (r *Room) broadcast(senderName, msg string) {
+func (r *Room) broadcast(senderName, msg string, isPluginMessage ...bool) {
 	if msg == "" {
 		return
 	}
+
+	// Now we know it is not a DM, so this is a safe place to add the hook for sending the event to plugins
+	if len(isPluginMessage) == 0 {
+		// this is the only way to make sure Plugins don't receive their own messages...
```
Arkaeriit, line 215: **question:** I don't really understand what you are doing to prevent plugins from receiving their own messages.

rpc.go
```diff
@@ -114,6 +116,19 @@ func (s *pluginServer) RegisterListener(stream pb.Plugin_RegisterListenerServer)
 			continue
 		}
 
+		if !isColorNames {
+			message.(*pb.Event).From = stripansi.Strip(message.(*pb.Event).From)
+			message.(*pb.Event).Msg = stripansi.Strip(message.(*pb.Event).Msg)
+		}
+
+		// This isn't a very nice solution to detect system messages
```
Arkaeriit, line 124: **thought:** I'm afraid there isn't any better solution, apart from refactoring broadcasting all over the codebase.

Arkaeriit:
Looks rather good to me. What did you do for testing?

PPTide:
I tested the the example plugin and made a simple plugin to just print the outputs into the terminal to test the two new parameters

main.go
```diff
@@ -205,10 +205,19 @@ func main() {
...
 	if msg == "" {
 		return
 	}
+
+	// Now we know it is not a DM, so this is a safe place to add the hook for sending the event to plugins
+	if len(isPluginMessage) == 0 {
+		// isPluginMessage is passed in when the message is broadcast by a plugin message
+		// we won't send that message to plugins b/c many plugins will go into an infinite loop reacting to their own message
+		msg = getMiddlewareResult(senderName, r.name, msg)
+		sendMessageToPlugins(msg, senderName, r.name)
```
Arkaeriit, line 218: **question:** If I understand correctly, this means that plugins receive no messages from any other plugins. Is that right?

Arkaeriit: Answered on Devzat: it is indeed so the comment is clear enough.

Arkaeriit:
Just rebase your MR on master so the dependencies updates are not duplicated and this will be ready to be merged.

PPTide:
pr with git fixed #253

PR #253, opened 2025-03-11
Rpc add options for bridge (fix git :c)

PPTide:
i think this fixes the bad git stuff xD #251


Arkaeriit:
You know you can push force, right? ;)

Arkaeriit:
Thanks for this big improvement!

PR #260, opened 2025-09-30
Update plugin dependency for new API.

Arkaeriit:
The latest updates to the API require the latest
plugin version. Indeed, the color fields were
needed but absent.


Arkaeriit:
As this is an important fix that is needed to use the go devzat API, I'll merge it without your review @quackduck.

Issue #238, opened 2024-08-16
joining without having ssh key results in scary ssh message


quackduck:
> ishan@devzat.hackclub.com: Permission denied (publickey).

options:

1. we should have a little message above this saying what the problem is (need pubkey gen)
2. we should allow joining but send a message from devbot encouraging pubkey gen soon.
3. ??? some other idea for how to not lose people who get scared by this message

2 seems better to me if possible.

Arkaeriit:
Option 2 would indeed be a good idea. Do you want me to work on it?

quackduck:
yeah can we look into whether this is technically feasible?

CaenJones:
You could set a custom ssh banner or MOTD

PR #278, opened 2026-02-11
Add an Interactive login type as a secondary login option

PPTide:
PR for #238.

Images of the interactive login in OpenSSH and Termius:
Screenshot 2026-02-11 at 11 36 31
Screenshot 2026-02-11 at 11 36 44

I'm open to changes to the wording of the text xD
We should also maybe add a command in Devzat (or the login screen) where the process of creating a ssh-key is explained.

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/acarl005/stripansi"
 	"github.com/gliderlabs/ssh"
 	terminal "github.com/quackduck/term"
+	cryptoSSH "golang.org/x/crypto/ssh"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


PPTide:
golangci-lint is broken not the formating of my code i think

PPTide:
@quackduck What are your thoughts? Ig this is a bigger change (allowing passkey-less login on the default :22 port aswell)

quackduck:
can you explain the idea a bit more? also i don't understand the second screenshot. what's the feature?

quackduck:
iirc one reason behavior differs between open and private devzat instances is because of the potential of spam and DOS. is this considered here?

edit: i may have misunderstood and possibly this is not a concern here. it's been a while since i thought about the design of devzat haha

PPTide:
> can you explain the idea a bit more? also i don't understand the second screenshot. what's the feature?

The second screenshot is just how the passkeyless login works in termius (an ssh client that is not only in the terminal). The first screenshot is openssh. 

I kinda also wanted to show that it's a different kind of login screen, that gets rendered differently by different ssh clients (ie i think putty also has a pop-up)

PPTide:
> iirc one reason behavior differs between open and private devzat instances is because of the potential of spam and DOS. is this considered here?
> 
> 
> 
> edit: i may have misunderstood and possibly this is not a concern here. it's been a while since i thought about the design of devzat haha

The passkeyless login only gets loaded when the Server is set to public. That is when before the Server on :443 would have started (which also doesn't require a passkey atm).

Now the Server at :443 is basically the same as the server on :22 and gets started if the altPort config is set.

PR #279, opened 2026-02-11
Fix Backlog messages always Default Messages

PPTide:
Backlog messages should be writen back with the same Message type as what they originally where. Atm join Messages in the Backlog are prefixed with ": "


PPTide:
https://github.com/quackduck/devzat/pull/278#issuecomment-3883647720

again

PR #281, opened 2026-02-11
Allow plugins to talk to each other

PPTide:
I still need to clean this up but just wanted to put this here already xD

Screenshot 2026-02-11 at 21 19 27

### DIFF
--- a/main.go
+++ b/main.go
@@ -218,11 +218,11 @@ func (r *Room) broadcast(msg Message) {
 	}
 
 	// Now we know it is not a DM, so this is a safe place to add the hook for sending the event to plugins
-	if msg.sendToPlugin {
+	if true {
 		// isPluginMessage is passed in when the message is broadcast by a plugin message
 		// we won't send that message to plugins b/c many plugins will go into an infinite loop reacting to their own message
-		msg.text = getMiddlewareResult(PluginMessage{msg, r.name})
-		sendMessageToPlugins(PluginMessage{msg, r.name})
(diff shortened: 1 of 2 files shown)


PPTide:
Just to add this adding support for commands would be slightly more annoying as the commands are handeled in a method that takes in the line and the user.

main.go
```diff
@@ -651,32 +651,32 @@ type Message struct {
...
+	return Message{"", text, NoSenderMessage, ""}
 }
 
 func NewPrivateMessage(senderName, text string, isSender bool) Message {
 	messageType := PrivateMessageReceive
 	if isSender {
 		messageType = PrivateMessageSend
 	}
-	return Message{senderName, text, messageType, true}
+	return Message{senderName, text, messageType, ""}
```
Arkaeriit, line 673: I really like how you used the new message struct to add more feature to it whilst keeping the high-level code just as readable. Nice job!

rpc.go
```diff
@@ -35,11 +35,16 @@ var (
...
-	ListenersMiddleware    = make([]chan pb.MiddlewareChannelMessage, 0, 4)
+	ListenersNonMiddleware = make([]channelAndToken[pb.MiddlewareChannelMessage], 0, 4)
+	ListenersMiddleware    = make([]channelAndToken[pb.MiddlewareChannelMessage], 0, 4)
 	Tokens                 = make(map[string]string, 10)
 )
 
+type channelAndToken[T any] struct {
+	channel chan T
+	token   string
+}
```
Arkaeriit: I feel like this struct's name should reflect more how it is used than what is inside. Maybe something like `listenerData` would be more suitable.

PR #282, opened 2026-02-12
Plugins can send commands

Arkaeriit:
This PR refactor how messages and commands are handled to let plugins execute Devzat commands. This has been tested with the courier plugin on another Devzat instance.

### DIFF
--- a/discord.go
+++ b/discord.go
@@ -154,7 +154,7 @@ func discordMessageHandler(_ *discordgo.Session, m *discordgo.MessageCreate) {
 			Log.Println("Overflow in Slack channel")
 		}
 	}
-	runCommands(msgContent, DiscordUser)
+	runCommands(NewMessage(DiscordUser, msgContent))
 }
 
 var cacheSize = 20
--- a/util.go
+++ b/util.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"math"
 	"math/rand"
+	"net"
 	"net/http"
 	"os"
 	"runtime/debug"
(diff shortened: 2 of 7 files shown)


util.go
```diff
@@ -504,3 +506,51 @@ func genKey() (ed25519.PrivateKey, ssh.PublicKey, error) {
 	}
 	return priv, sshPubKey, nil
 }
+
+func getMessageSenderName(msg Message) string {
```
PPTide: is there a reason this is a function and not a method on Message?

discord.go
```diff
@@ -154,7 +154,7 @@ func discordMessageHandler(_ *discordgo.Session, m *discordgo.MessageCreate) {
 			Log.Println("Overflow in Slack channel")
 		}
 	}
-	runCommands(msgContent, DiscordUser)
+	runCommands(NewMessage(DiscordUser, msgContent))
```
PPTide, line 157: Do you think we should get rid of DiscordUser? i think the fake user also makes sense here? or am i forgetting smth?

(Or get rid of discord/slack altogether? :p)

PPTide (changes requested):
it makes sense having the message have an actual user. Great!

util.go
Arkaeriit: Nope, too much C fried my brain and I forgot method were a thing ^^  
It would indeed make more sense to use a method here, I'll change that.

discord.go
Arkaeriit: I'm working on another PR to get rid of the bridges, so I wanted to only do minimal changes on the bridge code.

I think it would be better to keep those two changes is separate PRs, in case we ever need to revert one.

PPTide: that makes sense yea :3

util.go
Arkaeriit: I've fixed that.

Arkaeriit: Actually, not done. I've messed up a rebase at some point ^^

Arkaeriit: Re-fixed now.

PR #283, opened 2026-03-23
Enable Config from Env Variables

PPTide:
- [x] Refactor Integrations into Config struct
- [x] Add github.com/sethvargo/go-envconfig to parse enviroment variables
- [ ] Edit Readme to include these changes

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,27 +1,30 @@
 package main
 
 import (
+	"context"
 	"io"
 	"log"
 	"os"
 	"strconv"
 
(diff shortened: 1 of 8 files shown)


PPTide:
This is Regarding #275 btw I'll look at creating a good Docker container (and github workflow to make it) in a seperate PR.

I do think the point of having a docker container as a lot of people use docker to selfhost most of the Home Lab and thus adding a Docker container is easier then adding a go application (even tho go makes that very easy).

PR #284, opened 2026-05-12
Fix messages getting cut of weirdly

PPTide:
Messages had the possibility of getting cut of weirdly when the message included `-` and was `u.winWidth+range(1, dashCount)` characters long. For example: (Term width is 80)
```
Arkaeriit: Maybe there is a off-by-one error in the way the line width is counte
d.
```
Or:
```
Arkaeriit: Pretty good. I went to the catacombs and some people are re-painting 
a
           big room to make it look like a night club?
```
(Idk how @Arkaeriit manages to always hit the exact character count to overflow xD)

Both cases should have the last word (`counted.` or `a`) in the new line (with indentation)

This was b/c a [dependency of glamour](https://github.com/muesli/reflow) was counting the `-` character wrong. I updated glamour and the got rid of that dependency. (I also added tests :p) 

I don't know why my PR description is so long...（ ｉ _ ｉ ）


Arkaeriit:
Nice readable commits, thank you!

quackduck:
wow

PR #285, opened 2026-05-14
[WIP] Prototype of a username/password authentication system

plotfi:
* Adds require_auth and require_admin_auth config flags
* When require_auth is enabled, only registered users can connect and must provide a password
* Admins identified by SSH key hash bypass auth unless require_admin_auth is also set
* Passwords are bcrypt-hashed and stored in user-auth.json
* Prefs are keyed by authenticated username rather than IP-based ID
* Sensitive commands (adduser, removeuser, passwd) are intercepted before room broadcast to prevent password leakage

Full disclosure, I prototyped this using Claude. But I still think its straight forward enough to look at as inspiration for a username/password login system.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -72,6 +72,10 @@ var (
 		{"8ball", eightBallCMD, "`question`", "Always tells the truth."},
 		{"rmdir", rmdirCMD, "#`room`", "Remove an empty room"},
 		{"logs", logsCMD, "[`lines`]", "Show server logs (admin)"},
+		{"adduser", addUserCMD, "`username` `password`", "Register a user with a password (admin)"},
+		{"removeuser", removeUserCMD, "`username`", "Remove a user registration (admin)"},
+		{"listusers", listUsersCMD, "", "List registered users (admin)"},
+		{"passwd", passwdCMD, "", "Change your own password"},
 	}
 	SecretCMDs = []CMD{
 		{"ls", lsCMD, "???", "???"},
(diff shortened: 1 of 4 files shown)


PPTide:
Devzat already has an [allowlist](https://github.com/quackduck/devzat/blob/main/Admin's%20Manual.md) which enables similar functionality to this but using ssh keys. Adding password auth would be a major change to how Devzat works. 

It also honestly doesn't make any sense to have password auth and passkey auth and setting the server to public (to enable connecting w/o passkey on :443) is the opposite of your intentions. 

I would always suggest opening an issue with your idea before spending tokens to see if (AI Generation) of the Idea is even appreciated.

plotfi:
@PPTide Does the allowlist get hot-reloaded or does it require a restart? Also, does it prevent new users from signing on or does it merely reserve usernames? Sorry to bother if this all should be more obvious.

Arkaeriit:
> @PPTide Does the allowlist get hot-reloaded or does it require a restart? Also, does it prevent new users from signing on or does it merely reserve usernames? Sorry to bother if this all should be more obvious.

The allowlist doesn't get hot-reloaded. Maybe adding a command to let an admin do it could work.

PPTide:
> > @PPTide Does the allowlist get hot-reloaded or does it require a restart? Also, does it prevent new users from signing on or does it merely reserve usernames? Sorry to bother if this all should be more obvious.
> 
> The allowlist doesn't get hot-reloaded. Maybe adding a command to let an admin do it could work.

hmmm...

Yea we should prob put the allowlist in a separate file. I.e: 

permissions.yaml:
```
allowlist:
  3a3534a3d4727957a7e22c05d751d8921ef4a667973bd6d864a40c8ba39b9680: 'PPTide: github.com/PPTide'
admins:
  3a3534a3d4727957a7e22c05d751d8921ef4a667973bd6d864a40c8ba39b9680: 'PPTide: github.com/PPTide'
```

with the allowlist having a command and the admins maybe being able to just be hot-reloaded?

Arkaeriit:
I wonder if maybe the allowlist could be handled the same way as bans. Instead of being files being edited by the server admin, there could be a command `allow` that can be used by any Devzat admin to add someone to the allowlist.

As adding new admin is a pretty serious task, maybe keeping it cold-reloaded woudn't be too much of an issue.

PR #287, opened 2026-06-15
Remove Discrod bridge

Arkaeriit:
As PPTide made a [Discord bridge plugin](https://git.pptie.de/PPTide/devzatDiscord), there is no need to keep it embedded in Devzat.

If we remove the Slack bridge in a subsequent PR, we could remove the `isBridge` flag to make the code cleaner.


PPTide:
Do you want to add a link to my repo in the admin manual? https://github.com/PPTide/devzatDiscord (The github push mirror)

Arkaeriit:
> Do you want to add a link to my repo in the admin manual? https://github.com/PPTide/devzatDiscord (The github push mirror)

That's not a bad idea. Maybe we should create a section of the manual to list all known plugins and libraries around them. What do you think about that?

PPTide:
That seems like a pretty good idea!

Arkaeriit:
> That seems like a pretty good idea!

I added it in this PR.

Arkaeriit:
I'm merging, we can always revert if things go boom.

